PR #11, opened 2019-04-30
Enable our SpannerConnectionFactory for driver name 'spanner'

elefeint:
Nothing else would work unless factory discovery file is in META-INF, and `SpannerConnectionFactoryProvider` returns true, so this was the first step. The remaining boilerplate from the WIP will follow.

Also:
* added dependency on Project Reactor.
* added Assert class for very basic argument checking. Would be nice to not duplicate this in every driver, but Spring is too heavy of a required dependency. Guava may be appropriate if we need anything more complex.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/util/Assert.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/util/Assert.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner.r2dbc.util;
+
+import javax.annotation.Nullable;
+
+/**
+ * Lightweight assertion support.
+ */
+public class Assert {
+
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/util/Assert.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner.r2dbc.util;
+
+import reactor.util.annotation.Nullable;
+
+/**
+ * Lightweight assertion support.
```
meltsufin: `@author` everywhere?

src/main/java/com/google/cloud/spanner/r2dbc/util/Assert.java
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Checks that a specified object reference is not {@code null} and throws a customized
+   * {@link IllegalArgumentException} if it is.
+   *
+   * @param o       the object reference to check for nullity
+   * @param message informative message to be used in the event that an
+   * {@link IllegalArgumentException} is thrown
+   * @throws IllegalArgumentException if {@code o} is {@code null}
+   */
+  public static void requireNonNull(@Nullable Object o, String message) {
```
dmitry-s: I think it is better to return the value if it isn't null, that's the way it's implemented in r2dbc H2 (and I suspect the rest of the sample implementations)
```
public static <T> T requireNonNull(@Nullable T t, String message) {
        if (t == null) {
            throw new IllegalArgumentException(message);
        }

        return t;
}
```

dmitry-s: the reason is it allows you to do that:
```
this.name = Assert.requireNonNull(name, "name must not be null");
```
instead of
```
Assert.requireNonNull(name, "name must not be null")
this.name = name;
```

elefeint: Good point; I'll put the return value back.

PR #20, opened 2019-05-02
Results Class impl

### LINKED ISSUE #7: Implementation of Result API
Essentially creates Fluxes and Monos from client library results.

ChengyuanZhao:
fixes #7 

implementation of result class with unit tests


src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,83 @@
...
   }
 
   @Override
   public <T> Flux<T> map(BiFunction<Row, RowMetadata, ? extends T> f) {
-    return Flux.just(f.apply(null, null));
+    return Flux.fromIterable(() -> new Iterator<T>() {
+
+      private Boolean hasNext;
+
+      private Struct currentStruct;
```
elefeint: Are we going to use it outside of next()? It feels like it wants to be a local variable.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,83 @@
...
+
+      @Override
+      public T next() {
+        if (!hasNext()) {
+          throw new UnsupportedOperationException("This ResultSet has no more rows to get.");
+        }
+        this.currentStruct = SpannerResult.this.resultSet.getCurrentRowAsStruct();
+        this.hasNext = SpannerResult.this.resultSet.next();
+        return f
+            .apply(new SpannerRow(this.currentStruct), new SpannerRowMetadata(this.currentStruct));
```
elefeint: (non-actionable for this PR) do you recall if column metadata is available in GrpcStruct? When using Spanner API directly, the metadata came in only on the first PartialResultSet, and was retrievable as partialResultSet.getMetadata().

src/main/java/com/google/cloud/spanner/r2dbc/SpannerRow.java
```diff
@@ -16,15 +16,32 @@
...
+   * Constructor.
+   *
+   * @param struct the row from Cloud Spanner.
+   */
+  public SpannerRow(Struct struct) {
+    this.struct = struct;
+  }
 
-  public SpannerRow() {
+  @VisibleForTesting
```
elefeint: (non-actionable for this PR, also personal opinion) I am hoping to remove VisibleForTesting when we have implemented get().

src/test/java/com/google/cloud/spanner/r2dbc/SpannerResultTest.java
```diff
@@ -17,27 +17,94 @@
...
   }
 
+  @Test
+  public void mapErrorTest() throws InterruptedException {
+    try {
+      new SpannerResult(this.resultSet).map((row, metadata) -> {
+        throw new RuntimeException();
+      }).onErrorStop().collectList().block();
+    } catch (RuntimeException e) {
+      verify(this.resultSet, times(1)).close();
```
elefeint: This verify is likely never called. Could you check?

src/test/java/com/google/cloud/spanner/r2dbc/SpannerResultTest.java
```diff
@@ -17,27 +17,94 @@
...
+   */
+  @Before
+  public void setup() {
+    Struct struct1 = Struct.newBuilder().set("id").to(Value.string("key1")).build();
+    Struct struct2 = Struct.newBuilder().set("id").to(Value.string("key2")).build();
+
+    MockResults mockResults = new MockResults();
+    mockResults.structs = Arrays.asList(struct1, struct2);
+    this.resultSet = mock(ResultSet.class);
+    when(this.resultSet.next()).thenAnswer((invocation) -> mockResults.next());
```
elefeint: You may be able to avoid having a testing structure MockResults by using consecutive stubbing:
 when(this.resultSet.next()).thenReturn(true, true, false);
when(this.resultSet.getCurrentRowAsStruct()).thenReturn(struct1, struct2);

Ref: https://static.javadoc.io/org.mockito/mockito-core/2.27.0/org/mockito/Mockito.html#when-T-

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,83 @@
...
+  }
+
   @Override
   public Publisher<Integer> getRowsUpdated() {
-    return null;
+    return this.rowsUpdated;
   }
 
   @Override
   public <T> Flux<T> map(BiFunction<Row, RowMetadata, ? extends T> f) {
```
dzou: Probably need to do some nullchecks on `resultSet` variable or at least handle what happens if user calls map() on resultset that just contains # of rows updated.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,83 @@
...
-    return Flux.just(f.apply(null, null));
+    return Flux.fromIterable(() -> new Iterator<T>() {
+
```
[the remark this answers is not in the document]
ChengyuanZhao: I think that's sub-optimal because because you would need to pass in the `ResultSet` and the mapping function `f` (need to pass in one or both depending on how the class is extracted). Whereas being a small anonymous class lets us use both directly.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
ChengyuanZhao: Oh totally right !!!

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,82 @@
...
-    return Flux.just(f.apply(null, null));
+    return Flux.fromIterable(() -> new Iterator<Struct>() {
+
+      private Boolean hasNext;
+
+      private Struct currentStruct;
+
+      @Override
+      public boolean hasNext() {
+        if (this.hasNext == null) {
```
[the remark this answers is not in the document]
ChengyuanZhao: Hmm so I'm leaning towards keeping it as an anonymous class (which have no constructors) because this iterator needs both the `ResultSet` and the `f` mapping function. If it's anonymous it can use both directly in-line here, but if it's a separate class we would need more boilerplate to pass in those two things.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
ChengyuanZhao: Hmmm I don't remember off the top of my head. Actually the constructors for both `Row` and `RowMetadata` are basically placeholders for the purposes of this PR (I think they would be done as part of the mapping/conversion PR ? )

src/main/java/com/google/cloud/spanner/r2dbc/SpannerRow.java
ChengyuanZhao: Yes that's right. Since the mapping/converting work isn't done yet I just wanted to have this here so I could get the value out of this object for testing. They can be removed when we can get values from the row using the real `get` implementation.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerResultTest.java
ChengyuanZhao: It is indeed called (breakpoint stops there), but I can move it into a finally block so it's for sure called.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
ChengyuanZhao: Added test for the null case! 

However we don't need to worry about the result set just containing a count. The client lib doesn't give back result sets for DML operations. It just gives back the long directly.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,83 @@
...
+          this.hasNext = SpannerResult.this.resultSet.next();
+        }
+        return this.hasNext;
+      }
+
+      @Override
+      public T next() {
+        if (!hasNext()) {
```
dzou: I think there may be an issue with calling [this.resultSet.next()](https://googleapis.github.io/google-cloud-java/google-cloud-clients/apidocs/index.html) because it advances the resultSet to the next row. So you might be calling it too many times from the next() and hasNext() calls in the iterator; user might skip over several results.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerResultTest.java
ChengyuanZhao: Hmmm the problem with this is I'm not sure what sequence the thing actually calls those two functions. For example a "real" mocked resultset like i have no will skip straight to struct2 if next() is called twice, but if they're just separate mocks like that then there is no relationship between the two. I'm not sure if for our simple test if it wouldn't work, but I fear it might be a hard thing to remember when debugging later on.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
ChengyuanZhao: Could you give an example ? I thought the `next()` call being in the null-check if-statement in `hasNext()` prevents that from happening.

dzou: Ah sorry you're right; I missed that.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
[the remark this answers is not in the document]
dzou: Got it.

mvnw
```diff
@@ -0,0 +1,286 @@
+#!/bin/sh
```
meltsufin: Wrapper is still here.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,84 @@
...
   public <T> Flux<T> map(BiFunction<Row, RowMetadata, ? extends T> f) {
-    return Flux.just(f.apply(null, null));
+
+    if (resultSet == null) {
+      return Flux.empty();
+    }
+
+    return Flux.fromIterable(() -> new Iterator<T>() {
+
+      private Boolean hasNext;
```
meltsufin: Can you just change this to:
`private boolean hasNext = resultSet.next();`
and then `hasNext()` is much simpler.

meltsufin: Also, can the map function be called more than once? Then we might have a problem.

mvnw
ChengyuanZhao: TBH id why it's showing up in the diff ! It's in master already and this is diffed against master...

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
ChengyuanZhao: good idea ! changed. 

Regarding the calling map more than once on a result: that's a good point. Perhaps we should throw an error if it's called twice?

mvnw
elefeint: You might have to rebase against master, or manually delete the files in this branch.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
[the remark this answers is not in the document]
elefeint: We'll always have the option to refactor in the future if the logic grows any more complex (which it likely will, once backpressure is factored in).

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
elefeint: For the future, we may want to have different implementations of `SpannerQueryResultSet` vs `SpannerUpdateResultSet` or some such. There isn't much overlap in functionality if one gives a result set, and the other only a count.

ChengyuanZhao: Yes, but both need to implement the `Result` API anyway... (meaning the update result set would just have a number and would need an implementation of `map`)

src/test/java/com/google/cloud/spanner/r2dbc/SpannerResultTest.java
elefeint: My concern is that if some behavior in our map() implementation changes to eat the exception, this test will still pass, as resultset.close() will be called on successful completion.
Maybe add a false flag for exception thrown and set it in catch? Then verify that it's set.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
meltsufin: What do other drivers like MySQL do? Do they have the ability to reset the cursor to the beginning?

elefeint: Well, map() just returns a publisher. There is nothing wrong with acquiring a publisher multiple times. The interesting part is subscribing to the publisher from multiple places.

fromIterable [docs](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#fromIterable-java.lang.Iterable-) say it creates a new iterator for each subscriber; we might want to test what that means empirically for the spanner result set.

elefeint: Ah, I see the problem; we call "next()" on iterator creation, so multiple publisher acquisitions are actually a problem.

elefeint: I'd create a tracking issue for supporting multiple invocations of map(), and tackle it in a different PR.

ChengyuanZhao: I might be missing something here but I believe our hands are tied... The client lib's result set cannot be reset and we obviously do not want to cache the query results for a replay so I think that means we logically can only return an empty flux (what happens now) or throw an error on a second map attempt.

meltsufin: I think this is a good question to ask the R2DBC team for feedback.

dzou: There might be a way you can "replay" the stream or cache it:

Methods like flux.cache(); flux.replay() etc:
https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#cache--
https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#replay--

ChengyuanZhao: @dzou  yes of course we could cache pretty easily inside the SpannerResult too, but the point is that the result may be huge and we wouldn't' want to keep it around (which may be why the user wants to operate on the results one-at-a-time in the first place).

ChengyuanZhao: @mp911de Hey Mark, what's the policy or guideline on a `Result` being mapped over more than once?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,84 @@
...
+
+      @Override
+      public boolean hasNext() {
+        if (this.hasNext == null) {
+          this.hasNext = SpannerResult.this.resultSet.next();
```
mp911de: What happens in `resultSet.next()`? Do you initiate server communication (e.g. fetching from a cursor)?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
mp911de: Results are stream objects which can be consumed only once. More specifically: Rows are expected to be read from a stream source and so a `Row` can be consumed only once. Typically, there are buffers associated with a `Row` that get released after applying the mapping function.

Calling `Row.get(…)` is permitted multiple times to the same column.

ChengyuanZhao: Thank you Mark !

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,84 @@
...
+        if (!hasNext()) {
+          throw new UnsupportedOperationException("This ResultSet has no more rows to get.");
+        }
+        Struct currentStruct = SpannerResult.this.resultSet.getCurrentRowAsStruct();
+        this.hasNext = SpannerResult.this.resultSet.next();
+        return f
+            .apply(new SpannerRow(currentStruct), new SpannerRowMetadata(currentStruct));
+      }
+    }).doOnComplete(this.resultSet::close)
+        .doOnError(error -> this.resultSet.close())
```
mp911de: Is `.close()` a blocking call?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
ChengyuanZhao: Yes that's right. The client library's implementation takes care of making further calls to Cloud Spanner to resume where it left off.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
ChengyuanZhao: It is yes. I should wrap it up in `Mono<void>` right?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
mp911de: This does not work well. We're expecting non-blocking behavior. That means, either use futures or the `Observable` API for Spanner communication. Any blocking calls inside an R2DBC impl are forbidden because that keeps threads blocked.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
mp911de: This does not work because it's still a blocking call. `.close()` should either return a future or similar to avoid blocking calls entirely.

meltsufin: Mark, do you have examples in any of the drivers of dealing with blocking calls?

mp911de:
No, because all transport is non-blocking. 

What I’m trying to express is: Either your calls are non-blocking or you’re not building a reactive driver. R2DBC is end to end non-blocking. You have to use a non-blocking transport layer, either through Spanner‘s async API or the Observable API.

ChengyuanZhao:
@mp911de   Understood. I've changed this   result impl to now hold a Flux of rows as seen in the implementations for MSSQL and H2.

ChengyuanZhao:
@meltsufin @elefeint PTAL

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,59 @@
 
 package com.google.cloud.spanner.r2dbc;
 
+import com.google.cloud.spanner.Struct;
```
meltsufin: This class only exists in `google-cloud-spanner` library which we're not using. The diff here doesn't seem right. Can you try rebasing your branch?

ChengyuanZhao: oh that's right :( 

I don't think we're actually at the point of having our own "row" type. I think this PR might need to be put on hold until we do?

meltsufin: Right, we need to figure out how we will be representing rows.

elefeint: You could fall back to a collection of `com.google.protobuf.Value`, which is what `PartialResultSet` contains. But reassembling the rows is going to be a big excercise on its own.

ChengyuanZhao: hmmm but should we be doing that row-assembly elsewhere ?

elefeint: Has to be in between Statement and Result. To me Result looks like a more natural place for it, but it will likely be necessary to have a helper class do pure reassembly.

ChengyuanZhao: Wait isn't this clearly a responsibility of the client ? I thought the `publisher<partialresult>` was just a placeholder in `Client`.

elefeint: Oh, good point. Client's implementation of streaming query could do the conversion and produce a stream of proper rows.

ChengyuanZhao: Would it make sense to just use the `StructReader` interface from the client-lib ? (it seems silly to have two Google teams write two interfaces for a Spanner Row)

elefeint: The real implementation is in `GrpcResultSet`, which is package private. We could ask to relax visibility on all functionality related to parsing streaming partials, but then we run into the original issue with blocking calls (possibly, citation needed) and not exposing request(N).

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -32,17 +34,20 @@
...
 
   /**
    * Constructor for read-only query execution.
    *
-   * @param resultSet the underlying result from Cloud Spanner.
+   * @param resultRows the underlying result from Cloud Spanner.
    */
-  public SpannerResult(Flux<Struct> resultSet) {
-    this.resultSet = Assert.requireNonNull(resultSet, "A non-null flux of rows is required.");
+  public SpannerResult(Flux<List<Value>> resultRows, ResultSetMetadata rowMetadata) {
```
meltsufin: I think that `com.google.spanner.v1.StructType` which you get by calling `resultSetMetadata.getRowType()` corresponds better to row metadata. So I would probably rename this variable to `resultSetMetadata`.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -64,11 +70,11 @@ public SpannerResult(int rowsUpdated) {
...
 
-    if (this.resultSet == null) {
+    if (this.resultRows == null) {
       return Flux.empty();
     }
 
-    return this.resultSet.map(row -> f
-        .apply(new SpannerRow(row), new SpannerRowMetadata(row)));
+    return this.resultRows.map(row -> f
+        .apply(new SpannerRow(row, this.rowMetadata), new SpannerRowMetadata(this.rowMetadata)));
```
meltsufin: Here you need to pass `StructType` instead of  `ResultSetMetadata`.
In other words, `SpannerResult` uses `ResultSetMetadata` and `SpannerRow` uses `StructType`, which is obtained from `resultSetMetadata.getRowType()`.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerRow.java
```diff
@@ -16,34 +16,41 @@
...
 /**
  * {@link Row} implementation for Cloud Spanner.
  *
  */
 public class SpannerRow implements Row {
 
-  private final Struct struct;
+  private final List<Value> values;
+
+  private final ResultSetMetadata rowMetadata;
```
meltsufin: I think `StructType` makes more sense here.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
ChengyuanZhao: @meltsufin PTAL. changed to use `List<value>` and proto metadata.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerRow.java
ChengyuanZhao: Done. PTAL

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,70 @@
...
   @Override
   public <T> Flux<T> map(BiFunction<Row, RowMetadata, ? extends T> f) {
-    return Flux.just(f.apply(null, null));
+
+    if (this.resultRows == null) {
+      return Flux.empty();
+    }
+
+    return Flux.defer(() -> {
+      ResultSetMetadata resultSetMetadata = this.resultSetMetadata.block();
```
meltsufin: If you're doing this, why not just keep `ResultSetMetadata` without the `Mono` wrapper?

ChengyuanZhao: @mp911de

Hey Mark, I have a question regarding this line of code here. 

The situation for Spanner is that when the first row of data arrives it also comes with the row metadata for all subsequent rows. We have it packaged here as a flux of rows and a mono of the metadata. 

Here I have a block() call for that mono metadata inside the flux.defer returning the the flux of actual rows. My thought is that since the metadata arrives always with the first row, that this block() wouldn't actually ever wait for anything since the metadata would always already be available when the first row is available. Is this incorrect thinking?

mp911de: Calling `.block()` can cause harm because it uses internally blocking synchronization utilities therefore as rule: Don't ever use it (except on very rare occasion in tests even there use `StepVerifier` as much as possible).

The better approach is to have an `AtomicReference` (or better, `AtomicReferenceFieldUpdater` that is much more resource-friendly) that you use to get the metadata and store it once until a result completes (`compareAndSet(null, myMetadata)`).

The communication with Spanner is very likely synchronous that you have a single TCP connection over which you receive data. Therefore, you have an absolute message ordering so that results are emitted likely on the same thread. With that ordering, you know that a second record cannot get emitted before the first one and so the atomic utility helps you to maintain state. Probably even just a `volatile` field would be sufficient.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerResult.java
```diff
@@ -16,25 +16,66 @@
...
   }
 
   @Override
   public <T> Flux<T> map(BiFunction<Row, RowMetadata, ? extends T> f) {
-    return Flux.just(f.apply(null, null));
+
+    if (this.resultRows == null) {
+      return Flux.empty();
+    }
+    return Flux.combineLatest(this.resultSetMetadata, this.resultRows, (metadata, row) -> f
```
ChengyuanZhao: @mp911de 

Would using `Flux.combineLatest` to "combine" the mono.cache() and the flux be appropriate ? 

@meltsufin @elefeint  FYI

mp911de: That would work but you're paying with some additional overhead. Guess I should have explained it better that `volatile` is a performance-optimized approach in comparison to use operators.

meltsufin: I think maybe for now, we can just go with `Flux.combineLatest` and work on optimizations that make the code harder to understand at a later stage. WDYT?

ChengyuanZhao: I can make an issue to remember this optimization. For now I would prefer to get this in so we can build out the statement class and the client function to assemble the `List<Value>` @meltsufin

ChengyuanZhao: https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/issues/32

ChengyuanZhao: @elefeint PTAL

pom.xml
```diff
@@ -52,6 +52,12 @@
       <version>4.12</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
```
elefeint: sorry, you'll have to rebase or merge -- I just got #26 in, which also adds Mockito.

ChengyuanZhao: merged.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerResultTest.java
```diff
@@ -17,27 +17,70 @@
...
+import reactor.core.publisher.Mono;
 
 /**
  * Test for {@link SpannerResult}.
  */
 public class SpannerResultTest {
 
+  private Flux<List<Value>> resultSet;
+
+  private Mono<ResultSetMetadata> resultSetMetadata = Mono
```
elefeint: Mono.just(ResultSetMetadata.newBuilder().build())? It's reusable.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerResultTest.java
```diff
@@ -17,27 +17,70 @@
...
+        .just(Collections.singletonList(Value.newBuilder().setStringValue("key1").build()),
+            Collections.singletonList(Value.newBuilder().setStringValue("key2").build()));
+  }
+
+  @Test
+  public void getRowsUpdatedTest() {
+    assertThat(
+        ((Mono) new SpannerResult(this.resultSet, this.resultSetMetadata).getRowsUpdated()).block())
+        .isEqualTo(0);
+    assertThat(((Mono) new SpannerResult(Mono.fromSupplier(() -> 2)).getRowsUpdated()).block())
```
elefeint: Mono.just() instead of Mono.fromSupplier() here, too.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerResultTest.java
```diff
@@ -17,27 +17,70 @@
...
+    String metadataString = this.resultSetMetadata.block().toString();
+    assertThat(new SpannerResult(this.resultSet, this.resultSetMetadata).map((row, metadata) ->
+        ((SpannerRow) row).getValues().get(0).getStringValue() + "-"
+            + ((SpannerRowMetadata) metadata).getRowMetadata().toString()).collectList().block())
+        .containsExactly("key1-" + metadataString, "key2-" + metadataString);
   }
 
+  @Test
+  public void noResultsMapTest() {
+    assertThat(new SpannerResult(Mono.fromSupplier(() -> 2)).map((x, y) -> "unused"))
```
elefeint: Mono.just()

ChengyuanZhao: Somehow I forgot about that function! fixed.

PR #21, opened 2019-05-02
Add initial Transaction-related classes

dzou:
This PR adds the initial scaffolding for Transaction-related classes.

Includes: begin(), commit(), and close().

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
@@ -16,8 +16,9 @@
 
 package com.google.cloud.spanner.r2dbc;
 
-import com.google.cloud.spanner.DatabaseClient;
-import com.google.cloud.spanner.TransactionManager;
+import com.google.cloud.spanner.TransactionManager.TransactionState;
+import com.google.cloud.spanner.r2dbc.client.Client;
+import com.google.cloud.spanner.r2dbc.client.SpannerTransaction;
 import io.r2dbc.spi.Batch;
 import io.r2dbc.spi.Connection;
 import io.r2dbc.spi.IsolationLevel;
@@ -34,31 +35,31 @@ public class SpannerConnection implements Connection {
 
   private final Logger logger = LoggerFactory.getLogger(this.getClass());
 
-  private final DatabaseClient databaseClient;
+  private final Client client;
 
   private SpannerTransaction currentTransaction;
 
-  public SpannerConnection(DatabaseClient databaseClient) {
-    this.databaseClient = databaseClient;
+  public SpannerConnection(Client client) {
+    this.client = client;
     this.currentTransaction = null;
   }
 
   @Override
   public Publisher<Void> beginTransaction() {
-    return getTransactionState()
-        .flatMap(transactionState -> {
-          if (transactionState == TransactionManager.TransactionState.STARTED) {
+    return isInTransaction()
+        .flatMap(isInTransaction -> {
+          if (isInTransaction) {
             logger.debug(
                 "A transaction has already been started for this connection; "
                     + "beginTransaction() is a no-op.");
-            return Mono.just(currentTransaction);
+            return Mono.empty();
           } else {
-            return SpannerTransaction.startTransaction(databaseClient);
+            return client.startTransaction()
+                .doOnSuccess(newTransaction -> currentTransaction = newTransaction)
+                .then();
           }
         })
-        .doOnSuccess(spannerTransaction -> {
-          currentTransaction = spannerTransaction;
-        }).then();
+        .then();
   }
 
   @Override
@@ -72,11 +73,10 @@ public class SpannerConnection implements Connection {
 
   @Override
   public Publisher<Void> commitTransaction() {
-    return getTransactionState()
-        .flatMap(transactionState -> {
-          if (transactionState != TransactionManager.TransactionState.STARTED) {
-            logger.debug(
-                "Transaction is already in a terminal state; commitTransaction() is a no-op.");
+    return isInTransaction()
+        .flatMap(isInTransaction -> {
+          if (!isInTransaction) {
+            logger.debug("Transaction is not active; commitTransaction() is a no-op.");
             return Mono.empty();
           } else {
             return currentTransaction.commit();
@@ -119,12 +119,13 @@ public class SpannerConnection implements Connection {
     return null;
   }
 
-  private Mono<TransactionManager.TransactionState> getTransactionState() {
+  private Mono<Boolean> isInTransaction() {
     return Mono.defer(() -> {
       if (currentTransaction != null) {
-        return currentTransaction.getTransactionState();
+        return currentTransaction.getTransactionState()
+            .map(transactionState -> transactionState == TransactionState.STARTED);
       } else {
-        return Mono.just(TransactionManager.TransactionState.ABORTED);
+        return Mono.just(false);
       }
     });
   }
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
@@ -16,54 +16,115 @@
 
 package com.google.cloud.spanner.r2dbc;
 
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.TransactionManager;
 import io.r2dbc.spi.Batch;
 import io.r2dbc.spi.Connection;
 import io.r2dbc.spi.IsolationLevel;
 import io.r2dbc.spi.Statement;
+import java.util.Optional;
 import org.reactivestreams.Publisher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import reactor.core.publisher.Mono;
 
 /**
  * {@link Connection} implementation for Cloud Spanner.
  */
 public class SpannerConnection implements Connection {
 
+  private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+  private final DatabaseClient databaseClient;
+
+  public Optional<SpannerTransaction> currentTransaction = Optional.empty();
+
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -16,54 +16,115 @@
...
 /**
  * {@link Connection} implementation for Cloud Spanner.
  */
 public class SpannerConnection implements Connection {
 
+  private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+  private final DatabaseClient databaseClient;
+
+  private Optional<SpannerTransaction> currentTransaction = Optional.empty();
```
meltsufin: Please no optional in class fields.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -16,54 +16,115 @@
...
   public Publisher<Void> rollbackTransactionToSavepoint(String s) {
     return null;
   }
 
+  @Override
   public Publisher<Void> setTransactionIsolationLevel(IsolationLevel isolationLevel) {
     return null;
   }
+
+  private Mono<TransactionManager.TransactionState> getTransactionState() {
```
meltsufin: I checked H2 and MySQL implementations and they both use something like this instead:
```java
    private Mono<Void> useTransactionStatus(Function<TransactionStatus, Publisher<?>> function) {
        return Flux.defer(() -> function.apply(this.client.getTransactionStatus()))
            .then();
    }
```

I don't know if what you're doing is actually correct because the whole computation is not deferred, although it is a simple condition.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
dzou: I think this should be correct now; PTAL.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -16,54 +16,119 @@
...
+                "A transaction has already been started for this connection; "
+                    + "beginTransaction() is a no-op.");
+            return Mono.just(currentTransaction);
+          } else {
+            return SpannerTransaction.startTransaction(databaseClient);
+          }
+        })
+        .doOnSuccess(spannerTransaction -> {
+          if (currentTransaction != null) {
+            currentTransaction.close();
```
ChengyuanZhao: Is it the policy of the SPI to not let two transactions be open at the same time from the same connection? just wondering

dzou: Yeah good point; I think I will defer this to the user to avoid any unpredictable behavior.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -16,54 +16,119 @@
...
+            logger.debug(
+                "A transaction has already been started for this connection; "
+                    + "beginTransaction() is a no-op.");
+            return Mono.just(currentTransaction);
+          } else {
+            return SpannerTransaction.startTransaction(databaseClient);
+          }
+        })
+        .doOnSuccess(spannerTransaction -> {
+          if (currentTransaction != null) {
```
ChengyuanZhao: Is it possible that the `spannerTransaction` here the same exact one as `currentTransaction` from line 54?

dzou: Yeah you're right. I've removed the close() and just kept the assignment after the flatmap.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -16,54 +16,116 @@
...
   public Publisher<Void> setTransactionIsolationLevel(IsolationLevel isolationLevel) {
     return null;
   }
+
+  private Mono<TransactionManager.TransactionState> getTransactionState() {
+    return Mono.defer(() -> {
+      if (currentTransaction != null) {
+        return currentTransaction.getTransactionState();
+      } else {
+        return Mono.just(TransactionManager.TransactionState.ABORTED);
```
elefeint: Do we care about any internall Spanner transaction state, beyond "we are in a transaction" and "we are not in a transaction"?
I wonder if having a simple boolean would make sense instead of relying on internal spanner state.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -16,54 +16,116 @@
...
+            logger.debug(
+                "A transaction has already been started for this connection; "
+                    + "beginTransaction() is a no-op.");
+            return Mono.just(currentTransaction);
+          } else {
+            return SpannerTransaction.startTransaction(databaseClient);
+          }
+        })
+        .doOnSuccess(spannerTransaction -> {
+          currentTransaction = spannerTransaction;
```
elefeint: We only need to set `currentTransaction` when a new transaction is created. Perhaps instead of setting it in `doOnSuccess()`, we could set `currentTransaction` in the `else` clause above?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerTransaction.java
```diff
@@ -0,0 +1,61 @@
...
+  public Mono<Void> close() {
+    return Mono.fromRunnable(() -> transactionManager.close());
+  }
+
+  public Mono<Void> commit() {
+    return Mono.fromRunnable(() -> transactionManager.commit());
+  }
+
+  /** Starts a new Spanner transaction. */
+  public static Mono<SpannerTransaction> startTransaction(DatabaseClient databaseClient) {
```
elefeint: This static method will be hard to mock. I like the idea wrapper around low-level Spanner functionality, but then we might as well take full advantage of the extra layer of indirection and make it a mockable interface.
Also, could you create a sub-package "client" and move the wrapper there? It would be nice to keep the top-level package for SPI implementations only.
This will likely lead to two different classes with different concerns: a mockable interface `TransactionManager` (or `TransactionFactory` or similar), and a class to represent the current transaction containing the remaining methods. And the latter might well just be `TransactionContext` -- perhaps it does not need to be wrapped, if it's just a simple passthrough.

dzou:
@mp911de Hey Mark, was wondering if you could briefly look at this PR too in which we're starting to implement transactions and the R2DBC Connection interface. I saw you pointed out some pitfalls in other PR; any advice/tips you have would be appreciated as we are starting to learn more about Reactive!

src/main/java/com/google/cloud/spanner/r2dbc/SpannerTransaction.java
dzou: Done. I wrapped the `DatabaseClient` and moved to separate package; the database client is responsible for creating the transaction so it has the role of "transactionFactory" and I maintained the `SpannerTransaction` wrapper for the transactionContext and transactionManager objects from the client library.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
dzou: Good call, done.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
dzou: Done; made into boolean for simplicity.

src/main/java/com/google/cloud/spanner/r2dbc/client/SpannerClient.java
```diff
@@ -0,0 +1,43 @@
...
+  private final DatabaseClient databaseClient;
+
+  public SpannerClient(DatabaseClient databaseClient) {
+    this.databaseClient = databaseClient;
+  }
+
+  @Override
+  public Mono<SpannerTransaction> startTransaction() {
+    return Mono.fromSupplier(() -> {
+      TransactionManager transactionManager = databaseClient.transactionManager();
```
mp911de: For these methods, it would make use to bridge `ListenableFuture` (`SpannerGrpc.newFutureStub(channel).beginTransaction()`) as `Mono` has no real need for backpressure.

src/main/java/com/google/cloud/spanner/r2dbc/client/SpannerTransaction.java
```diff
@@ -0,0 +1,52 @@
...
+    this.transactionManager = transactionManager;
+    this.transactionContext = transactionContext;
+  }
+
+  public Mono<TransactionManager.TransactionState> getTransactionState() {
+    return Mono.fromSupplier(() -> transactionManager.getState());
+  }
+
+  public Mono<Void> rollback() {
+    return Mono.fromRunnable(() -> transactionManager.rollback());
```
mp911de: These methods should use the async API, not the blocking one.

elefeint: Thank you for the review!
Turning the calls into asynchronous ones is actually on the client library team's roadmap. We have two pretty good choices here:
1) work with client library team to turn these simple API methods to be asynchronous faster, then make use of them.
2) use a hybrid approach, relying on client library for the streaming querying, while directly using Spanner API over gRPC for the other (commit/rollback/start session) API methods. 

I'll go figure out how feasible/fast (1) will be now...

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
mp911de: Please note: H2 does not involve any I/O layer so calls remain there in the same Thread so H2 can behave sometimes different. Postgres, SQL Server, and MySQL drivers are a better reference.

PR #23, opened 2019-05-06
Switch to gRPC client for Cloud Spanner

meltsufin:
Also introduces the Client abstraction to hide interactions with the
gRPC Spanner service, and implements the basic executeStreamingSql
method wrapping the result into a Reactive Streams
Publisher<PartialResultSet>.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner.r2dbc.client;
+
+import com.google.spanner.v1.ExecuteSqlRequest;
+import com.google.spanner.v1.PartialResultSet;
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Mono;
+
+/**
+ * An abstraction that wraps interaction with the Cloud Spanner Database APIs.
+ */
+public interface Client {
+
+  /**
+   * Release any resources held by the {@link Client}.
+   *
+   * @return a {@link Mono} that indicates that a client has been closed
+   */
+  Mono<Void> close();
+
+  /**
+   * Execute a streaming query and get partial results.
+   */
+  Publisher<PartialResultSet> executeStreamingSql(ExecuteSqlRequest request);
+
+}
--- a/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
@@ -37,6 +37,7 @@ import reactor.core.publisher.Mono;
  */
 public class GrpcClient implements Client {
 
+  public static final String HOST = "spanner.googleapis.com";
   public static final int PORT = 443;
 
   private final SpannerStub spanner;
@@ -47,7 +48,7 @@ public class GrpcClient implements Client {
   public GrpcClient() throws IOException {
     // Create a channel
     ManagedChannel channel = ManagedChannelBuilder
-        .forAddress("spanner.googleapis.com", PORT)
+        .forAddress(HOST, PORT)
         .build();
 
     // Create blocking and async stubs using the channel
--- a/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2019 Google LLC
+ *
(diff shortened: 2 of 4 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
```diff
@@ -0,0 +1,41 @@
...
+   * Release any resources held by the {@link Client}.
+   *
+   * @return a {@link Mono} that indicates that a client has been closed
+   */
+  Mono<Void> close();
+
+  /**
+   * Execute a streaming query and get partial results.
+   */
+  Publisher<PartialResultSet> executeStreamingSql(ExecuteSqlRequest request);
```
dzou: Maybe hold off on adding this method to the interface for now? Because I feel that the interface should be GRPC/client-library agnostic; therefore the param and return type of the method will probably need to be changed later to not be GRPC-specific classes.

ChengyuanZhao: I think this should be `Publisher<Struct>`. It makes sense that the client produces units of rows (structs) instead of partial results that don't correspond to rows directly.  See the `ReactorNettyClient` for MSSQL which produces a `Flux<Message>` where each message is a row.

src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
```diff
@@ -0,0 +1,41 @@
...
+ * An abstraction that wraps interaction with the Cloud Spanner Database APIs.
+ */
+public interface Client {
+
+  /**
+   * Release any resources held by the {@link Client}.
+   *
+   * @return a {@link Mono} that indicates that a client has been closed
+   */
+  Mono<Void> close();
```
ChengyuanZhao: Actually if the result of the following function (`Flux/Publisher<Struct>`) is what it is, then that implies the closing and resource management is handled within the implementation of that object rather than exposed here, right?

src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
elefeint: Should we turn it into a Publisher<Row> right away? It ought to be the client's problem how to go from gRPC-specifric abstractions into an R2DBC suitable one.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -0,0 +1,107 @@
...
+
+  private final SpannerStub spanner;
+
+  /**
+   * Initializes the Cloud Spanner gRPC async stub.
+   */
+  public GrpcClient() throws IOException {
+    // Create a channel
+    ManagedChannel channel = ManagedChannelBuilder
+        .forAddress("spanner.googleapis.com", PORT)
```
elefeint: Could you make it a constant, similar to `PORT`?

src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
meltsufin: This is certainly just a placeholder.

I think this is the purest form we can provide that is truly reactive, and it can be useful to someone dropping down to this level when they're dealing with extremely large column values or rows.

I do like the `Publisher<Struct>` idea, but I can't implement it yet. :smile: 
Let me move this method to a private one for now, and replace it with `Publisher<Struct>`.

src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
meltsufin: The `Client` creates `SpannerStub` which I thought might need to be closed. It doesn't seen to be the case, but I left the method on the interface just as a stub because we'll probably have things to close here in the future as this class evolves.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -0,0 +1,107 @@
...
+        .withCallCredentials(callCredentials);
+  }
+
+  @Override
+  public Mono<Void> close() {
+    return null;
+  }
+
+  @Override
+  public Publisher<PartialResultSet> executeStreamingSql(ExecuteSqlRequest request) {
```
elefeint: Since this functionality is here, could you add a test?

src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
dzou: Sure placeholder sounds good; something that we can solidify at a later point in time.

elefeint: I meant a publisher of Rows. But yes, let's get it in and evolve later.

meltsufin: We actually don't have a `Struct` class in the gRPC Spanner library. We may need to use something like `ListValue`?

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
meltsufin: Yes, a test makes sense, but since it would be somewhat involved, and I would like to unblock everyone ASAP, can we do it in a separate PR?

elefeint: yep

mp911de: That code looks way more appropriate. I suggest using Project Reactor's utilities (such as `Flux.create(sink -> …)`) to:

1. Not implement `Publisher` yourself
2. Benefit from Reactor's subscriber `Context` (See `CoreSubscriber`) as Spring heavily uses the `Context` feature to bind resources (e.g. transactions) to a subscription.

Query methods are good candidates to be tested via the Reactive Streams TCK (see https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck).

PR #25, opened 2019-05-06
Change ExecuteStreamingSql style to use Flux.create

dzou:
As per [Mark's comment](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/23#discussion_r281350638), modifies the executeStreamingSql method to use `Flux.create` instead of creating Publisher manually.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
@@ -29,7 +29,7 @@ import io.grpc.stub.ClientCallStreamObserver;
 import io.grpc.stub.ClientResponseObserver;
 import java.io.IOException;
 import org.reactivestreams.Publisher;
-import org.reactivestreams.Subscription;
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -67,42 +67,30 @@ public GrpcClient() throws IOException {
...
-          @Override
-          public void beforeStart(
-              ClientCallStreamObserver<ExecuteSqlRequest> clientCallStreamObserver) {
-
-            clientCallStreamObserver.disableAutoInboundFlowControl();
-
-            subscriber.onSubscribe(new Subscription() {
-              @Override
-              public void request(long l) {
-                clientCallStreamObserver.request((int) l);
```
meltsufin: This and the `cancel` doesn't seem to have made it over to the `Flux` version.

dzou: I don't think these need to be implemented for the `Flux.create` version.

dzou: Ok, restored the equivalent code; PTAL.

Issue #14, opened 2019-05-01
Add integration test for reading and writing from real Spanner instance


elefeint:
Something more polished than the [wip](https://github.com/elefeint/r2dbc-wip/blob/49cd39afccf7f83cb0c1682611cadcef517f884d/src/test/java/io/r2dbc/spanner/SampleSpannerTest.java#L86).

ChengyuanZhao:
@elefeint  I was thinking of making a large table with columns of each type (including the array types) and then generating many rows of random data and doing LARGE reads so we can see many partial result sets in the result with chunks and then also trying out some insert /update /delete statements .

elefeint:
Bonus points for data not being random, for easier troubleshooting.

dzou:
Doing the large integration test will be useful for verifying our assumptions about partial result set contents. For example, I learned that the stats are actually provided in the last element emitted in the stream, not the first: https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/issues/75#issuecomment-495764310

PR #26, opened 2019-05-07
create R2DBC connection as a Spanner session created over gRPC

elefeint:
This change:
* Creates a real `Connection` object with minimum necessary settings.
* When a new Connection is created, it creates a new Spanner session and keeps its name. The session name is necessary for most future operations.
* Added a utility `ObservableReactiveUtil` that takes care of the boilerplate of converting from `StreamObservable` into a `Mono`.
* There is a new integration test, gated with `it.spanner`. You'll have to create a new instance "reactivetest" and a new database "testdb" to run it with no errors. The proper integration setup will be done in #14 .
* I added a _test-only_ dependency on `com.google.cloud:google-cloud-core` to take advantage of `ServiceOptions` project ID autodiscovery. Edit: decided to keep it as is for now; possibly remove completely in the future.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,9 +16,11 @@
     <maven.compiler.target>1.8</maven.compiler.target>
     <r2dbc.version>1.0.0.M7</r2dbc.version>
     <reactor.version>Californium-SR6</reactor.version>
-    <google-cloud-java.version>0.90.0-alpha</google-cloud-java.version>
-    <google-auth.version>0.11.0</google-auth.version>
+    <google-auth.version>0.15.0</google-auth.version>
     <grpc.version>1.20.0</grpc.version>
+    <mockito.version>2.23.0</mockito.version>
+    <grpc-spanner.version>1.17.0</grpc-spanner.version>
+    <google-cloud-core.version>1.72.0</google-cloud-core.version>
   </properties>
 
 
@@ -38,6 +40,7 @@
     <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>grpc-google-cloud-spanner-v1</artifactId>
+      <version>${grpc-spanner.version}</version>
     </dependency>
 
     <dependency>
@@ -49,7 +52,11 @@
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-auth</artifactId>
-      <version>${grpc.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty-shaded</artifactId>
     </dependency>
 
     <!-- test dependencies -->
@@ -65,22 +72,38 @@
       <version>4.12</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>${mockito.version}</version>
+      <scope>test</scope>
+    </dependency>
+
+    <!-- TODO: this is a test dependency for now. Should project ID discovery be part of the driver itself? -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-core</artifactId>
+      <version>${google-cloud-core.version}</version>
+      <scope>test</scope>
+    </dependency>
 
   </dependencies>
 
   <dependencyManagement>
     <dependencies>
+
+      <dependency>
+        <groupId>io.grpc</groupId>
+        <artifactId>grpc-bom</artifactId>
+        <version>${grpc.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+
       <dependency>
         <groupId>io.projectreactor</groupId>
         <artifactId>reactor-bom</artifactId>
         <version>${reactor.version}</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
-      <dependency>
-        <groupId>com.google.cloud</groupId>
-        <artifactId>google-cloud-bom</artifactId>
-        <version>${google-cloud-java.version}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
@@ -23,14 +23,13 @@ import io.r2dbc.spi.Connection;
 import io.r2dbc.spi.IsolationLevel;
 import io.r2dbc.spi.Statement;
 import org.reactivestreams.Publisher;
-import reactor.core.publisher.Mono;
 
 /**
  * {@link Connection} implementation for Cloud Spanner.
  */
 public class SpannerConnection implements Connection {
 
-  private Mono<Session> session;
+  private Session session;
 
   private Client client;
 
@@ -39,7 +38,7 @@ public class SpannerConnection implements Connection {
    * @param client client controlling low-level Spanner operations
    * @param session Spanner session to use for all interactions on this connection.
    */
-  public SpannerConnection(Client client, Mono<Session> session) {
+  public SpannerConnection(Client client, Session session) {
 
     this.client = client;
 
@@ -86,7 +85,12 @@ public class SpannerConnection implements Connection {
     return null;
   }
 
-  public Mono<String> getSessionName() {
-    return this.session.map(s -> s.getName());
+  /**
+   * Returns the Spanner session associated with the current {@link Connection}.
+   * @return spanner session proto
+   */
+  public Session getSession() {
+    return this.session;
   }
+
 }
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
@@ -16,16 +16,38 @@
 
 package com.google.cloud.spanner.r2dbc;
 
+import com.google.cloud.spanner.r2dbc.client.Client;
+import com.google.spanner.v1.Session;
 import io.r2dbc.spi.Batch;
 import io.r2dbc.spi.Connection;
 import io.r2dbc.spi.IsolationLevel;
 import io.r2dbc.spi.Statement;
 import org.reactivestreams.Publisher;
+import reactor.core.publisher.Mono;
 
 /**
  * {@link Connection} implementation for Cloud Spanner.
  */
 public class SpannerConnection implements Connection {
+
+  private SpannerConnectionConfiguration config;
+
+  private Mono<Session> session;
+
+  private Client client;
+
+  /**
+   * Instantiates a Spanner session with given configuration.
+   * @param client client controlling low-level Spanner operations
+   * @param config database connection settings
+   */
+  public SpannerConnection(Client client, SpannerConnectionConfiguration config) {
+    this.config = config;
+
+    this.client = client;
+
+    this.session = this.client.createSession(config.getFullyQualifiedDatabaseName());
+  }
 
   public Publisher<Void> beginTransaction() {
     return null;
@@ -65,5 +87,9 @@ public class SpannerConnection implements Connection {
 
   public Publisher<Void> setTransactionIsolationLevel(IsolationLevel isolationLevel) {
     return null;
+  }
+
+  public Mono<String> getSessionName() {
+    return this.session.map(s -> s.getName());
   }
 }
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
@@ -30,8 +30,6 @@ import reactor.core.publisher.Mono;
  */
 public class SpannerConnection implements Connection {
 
-  private SpannerConnectionConfiguration config;
-
   private Mono<Session> session;
 
   private Client client;
@@ -39,14 +37,13 @@ public class SpannerConnection implements Connection {
   /**
    * Instantiates a Spanner session with given configuration.
    * @param client client controlling low-level Spanner operations
-   * @param config database connection settings
+   * @param session Spanner session to use for all interactions on this connection.
    */
-  public SpannerConnection(Client client, SpannerConnectionConfiguration config) {
-    this.config = config;
+  public SpannerConnection(Client client, Mono<Session> session) {
 
     this.client = client;
 
-    this.session = this.client.createSession(config.getFullyQualifiedDatabaseName());
+    this.session = session;
   }
 
   public Publisher<Void> beginTransaction() {
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
@@ -23,20 +23,25 @@ import com.google.cloud.spanner.r2dbc.util.Assert;
  */
 public class SpannerConnectionConfiguration {
 
-  private String projectId;
+  private static final String FULLY_QUALIFIED_DB_NAME_PATTERN
+      = "projects/%s/instances/%s/databases/%s";
 
-  private String instanceName;
+  private final String projectId;
 
-  private String databaseName;
+  private final String instanceName;
+
+  private final String databaseName;
+
+  private final String fullyQualifiedDbName;
 
   /**
    * Basic property initializing constructor.
    *
+   * @param projectId GCP project that contains the database.
    * @param instanceName instance to connect to
    * @param databaseName database to connect to.
    */
-  // TODO: set up builder.
-  public SpannerConnectionConfiguration(
+  private SpannerConnectionConfiguration(
       String projectId, String instanceName, String databaseName) {
     this.projectId
         = Assert.requireNonNull(projectId, "projectId must not be null");
@@ -44,6 +49,9 @@ public class SpannerConnectionConfiguration {
         = Assert.requireNonNull(instanceName, "instanceName must not be null");
     this.databaseName
         = Assert.requireNonNull(databaseName, "databaseName must not be null");
+
+    this.fullyQualifiedDbName = String.format(
+        FULLY_QUALIFIED_DB_NAME_PATTERN, this.projectId, this.instanceName, this.databaseName);
   }
 
   public String getInstanceName() {
@@ -59,16 +67,41 @@ public class SpannerConnectionConfiguration {
   }
 
   /**
-   * Formats configuration properties into a fully qualified database name.
+   * Turns configuration properties into a fully qualified database name.
    * @return fully qualified database name
    */
   public String getFullyQualifiedDatabaseName() {
-    return "projects/"
-        + this.projectId
-        + "/instances/"
-        + this.instanceName
-        + "/databases/"
-        + this.databaseName;
+    return this.fullyQualifiedDbName;
+  }
+
+  public static class Builder {
+
+    private String projectId;
+
+    private String instanceName;
+
+    private String databaseName;
+
+    public Builder setProjectId(String projectId) {
+      this.projectId = projectId;
+      return this;
+    }
+
+    public Builder setInstanceName(String instanceName) {
+      this.instanceName = instanceName;
+      return this;
+    }
+
+    public Builder setDatabaseName(String databaseName) {
+      this.databaseName = databaseName;
+      return this;
+    }
+
+    public SpannerConnectionConfiguration build() {
+      return new SpannerConnectionConfiguration(
+          this.projectId, this.instanceName, this.databaseName);
+    }
+
   }
 
 }
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
@@ -26,12 +26,6 @@ public class SpannerConnectionConfiguration {
   private static final String FULLY_QUALIFIED_DB_NAME_PATTERN
       = "projects/%s/instances/%s/databases/%s";
 
-  private final String projectId;
-
-  private final String instanceName;
-
-  private final String databaseName;
-
   private final String fullyQualifiedDbName;
 
   /**
@@ -43,27 +37,13 @@ public class SpannerConnectionConfiguration {
    */
   private SpannerConnectionConfiguration(
       String projectId, String instanceName, String databaseName) {
-    this.projectId
-        = Assert.requireNonNull(projectId, "projectId must not be null");
-    this.instanceName
-        = Assert.requireNonNull(instanceName, "instanceName must not be null");
-    this.databaseName
-        = Assert.requireNonNull(databaseName, "databaseName must not be null");
+
+    Assert.requireNonNull(projectId, "projectId must not be null");
+    Assert.requireNonNull(instanceName, "instanceName must not be null");
+    Assert.requireNonNull(databaseName, "databaseName must not be null");
 
     this.fullyQualifiedDbName = String.format(
-        FULLY_QUALIFIED_DB_NAME_PATTERN, this.projectId, this.instanceName, this.databaseName);
-  }
-
-  public String getInstanceName() {
-    return instanceName;
-  }
-
-  public String getDatabaseName() {
-    return databaseName;
-  }
-
-  public String getProjectId() {
-    return projectId;
+        FULLY_QUALIFIED_DB_NAME_PATTERN, projectId, instanceName, databaseName);
   }
 
   /**
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
@@ -23,6 +23,8 @@ import com.google.cloud.spanner.r2dbc.util.Assert;
  */
 public class SpannerConnectionConfiguration {
 
+  private String projectId;
+
   private String instanceName;
 
   private String databaseName;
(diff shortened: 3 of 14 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -40,26 +43,39 @@
...
-  private final SpannerStub spanner;
+  private SpannerStub spanner;
 
   /**
-   * Initializes the Cloud Spanner gRPC async stub.
+   * Creates the gRPC stub with authentication.
+   * @throws IOException if credentials could not be acquired
    */
-  public GrpcClient() throws IOException {
+  public void initialize() throws IOException {
```
dzou: To avoid the checked exception, I think you can keep the constructor but make the constructor with signature:

`public GrpcClient(CallCredentials callCredentials)`

Then delegate the providing of CallCredentials to the SpannerConnectionFactory (i.e. something like `SpannerConnectionFactory.createClient(Credentials c)`. I think this approach makes sense since it allows the factory to create GrpcClient via default credentials or custom credentials via service account depending on what the user chooses. And keeps constructor free from IO.

Not too big of a fan of initialize() methods simply b/c you might forget to call it.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactory.java
```diff
@@ -30,19 +32,32 @@
...
 
   @Override
-  public Publisher<? extends Connection> create() {
-
-    return Mono.just(new SpannerConnection());
+  public Publisher<SpannerConnection> create() {
+
+    return Mono.defer(() -> {
+      try {
+        this.client.initialize();
```
ChengyuanZhao: Does `this.client` need to be initialized for every connection created? or just the first one ?

pom.xml
```diff
@@ -65,22 +72,38 @@
...
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>${mockito.version}</version>
+      <scope>test</scope>
+    </dependency>
+
+    <!-- TODO: this is a test dependency for now. Should project ID discovery be part of the driver itself? -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-core</artifactId>
```
dzou: I'm in favor of just keeping this scoped to test for now. I think it's good at test scope to make tests easier to write.

At the moment it might be simpler to leave it out of the driver release just to keep it smaller/light-weight; later down the road when we see common use-cases we could identify a need for it and add it back if necessary.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerIntegrationTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Before
+  public void enableIntegrationTest() {
+    assumeThat("Enable integration tests with -Dit.spanner=true.",
+        System.getProperty("it.spanner"), is("true"));
+  }
+
+  @Test
+  public void testCreatingSession() {
+
```
dzou: nit: rmv 2 newlines

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -49,4 +54,21 @@ public String getDatabaseName() {
...
+  public String getProjectId() {
+    return projectId;
+  }
+
+  /**
+   * Formats configuration properties into a fully qualified database name.
+   * @return fully qualified database name
+   */
+  public String getFullyQualifiedDatabaseName() {
+    return "projects/"
```
dzou: Perhaps this string could be built and cached at constructor level.

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Invokes a lambda that in turn issues a remote call, directing the response to a {@link Mono}
+   * stream.
+   * @param remoteCall lambda capable of invoking the correct remote call, making use of the
+   * {@link Mono}-converting {@link StreamObserver} implementation.
+   * @param <ResponseT> type of remote call response
+   * @return
+   */
+  public static <ResponseT> Mono<ResponseT> unaryCall(
```
ChengyuanZhao: this function is the key to it all ! nice !

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -16,17 +16,39 @@
...
 /**
  * {@link Connection} implementation for Cloud Spanner.
  */
 public class SpannerConnection implements Connection {
 
+  private SpannerConnectionConfiguration config;
+
+  private Mono<Session> session;
+
+  private Client client;
```
ChengyuanZhao: Why does the connection hold both a session and a client? A client spans the lives of many sessions, and if a SpannerConnection is basically a single session then it seems like the client shouldn't belong here.

dzou: I think you need the client to perform operations like beginTransaction and endTransaction; and in the grpc `beginTransaction` call you will need to pass in the session id.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -16,17 +16,39 @@
...
+import reactor.core.publisher.Mono;
 
 /**
  * {@link Connection} implementation for Cloud Spanner.
  */
 public class SpannerConnection implements Connection {
 
+  private SpannerConnectionConfiguration config;
+
+  private Mono<Session> session;
```
ChengyuanZhao: Do you think `Session` is the correct type to make available in `SpannerConnection`? 

I took a  look and it's got pretty low-level functions and i didn't see anything that looked familiar like a `TransactionContext` from the client lib that would facilitate these very high-level SQL-related methods we need to implement in here.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
ChengyuanZhao: Should those actions be packaged together with a session in some wrapper class then (perhaps as some runnable actions etc..) ? It just feels weird to have something that's essentially a session factory along with a session here

elefeint: Client is our bridge to all Spanner APIs, which are flat (so, CreateSession and CommitTransaction are both Spanner API methods, even though they deal with different object domains). We could certainly split the Client class into tighter abstractions if it grows unwieldy.

The Session proto is just a data holding proto. The one important piece of information is the *session name* -- it is necessary for all subsequent operations, such as [begin a transaction](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.BeginTransactionRequest) and [ExecuteStreamingSql](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Spanner.ExecuteStreamingSql) (session is mandatory in [ExecuteSqlRequest](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.ExecuteSqlRequest)).

The client library abstracted those low level concerns for us, but now that we are targeting gRPC API, we have a lower level set of operations to build from.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
elefeint: Yeah, I'll do that, and add a builder while I am there.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactory.java
elefeint: I think I'll take Dan's suggestion of having a credentials provider injected. Then we don't need the `initialize` method at all.
It's only needed the first time, for creating stubs.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
elefeint: I like the injectable credentials provider. It will be more testable, too.

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
elefeint: Yep, it should serve for all API calls except for `ExecuteStreamingSql`, which results in a Flux.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -16,17 +16,39 @@
...
+
+  private Client client;
+
+  /**
+   * Instantiates a Spanner session with given configuration.
+   * @param client client controlling low-level Spanner operations
+   * @param config database connection settings
+   */
+  public SpannerConnection(Client client, SpannerConnectionConfiguration config) {
```
meltsufin: I'm not sure if we need to pass the`SpannerConnectionConfiguration` here. Maybe just pass the `Session` instead?
That way only the connection factory needs to deal with `SpannerConnectionConfiguration`.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -66,4 +88,8 @@ public Statement createStatement(String sql) {
   public Publisher<Void> setTransactionIsolationLevel(IsolationLevel isolationLevel) {
     return null;
   }
+
+  public Mono<String> getSessionName() {
```
meltsufin: What's the purpose of this method?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -33,12 +35,15 @@
    * @param instanceName instance to connect to
```
meltsufin: document project id

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -33,12 +35,15 @@
...
+      String projectId, String instanceName, String databaseName) {
+    this.projectId
+        = Assert.requireNonNull(projectId, "projectId must not be null");
+    this.instanceName
+        = Assert.requireNonNull(instanceName, "instanceName must not be null");
+    this.databaseName
+        = Assert.requireNonNull(databaseName, "databaseName must not be null");
   }
 
   public String getInstanceName() {
```
meltsufin: Do we need this method?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -49,4 +54,21 @@ public String getDatabaseName() {
     return databaseName;
   }
 
+  public String getProjectId() {
```
meltsufin: Do we need this method?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactory.java
```diff
@@ -30,19 +32,32 @@
 
   private SpannerConnectionConfiguration config;
 
+  private Client client = new GrpcClient();
```
meltsufin: This will become a problem for testing. Please pass it through the constructor.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactory.java
```diff
@@ -30,19 +32,32 @@
...
+      }
+    });
   }
 
   @Override
   public ConnectionFactoryMetadata getMetadata() {
     return SpannerConnectionFactoryMetadata.INSTANCE;
   }
 
+  public void setClient(Client client) {
```
meltsufin: I don't think we need this, if the client comes in through the constructor.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
```diff
@@ -34,9 +36,20 @@
   /** R2DBC driver name for Google Cloud Spanner. */
   public static final String DRIVER_NAME = "spanner";
 
+  /** Option name for GCP Project. */
+  public static final Option<String> OPTION_PROJECT = Option.valueOf("project");
```
meltsufin: I don't see the convention in other options to prefix with `OPTION_`.

src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
```diff
@@ -26,6 +28,13 @@
  */
 public interface Client {
 
+  /**
+   * Prepares the client for performing Spanner operations.
+   * @throws IOException if credentials could not be acquired
+   */
+  public void initialize() throws IOException;
```
meltsufin: If anything, this should be done in the constructor.

src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
```diff
@@ -38,4 +47,9 @@
    */
   Publisher<PartialResultSet> executeStreamingSql(ExecuteSqlRequest request);
 
+  /**
+   * Create a Spanner session.
+   */
+  Mono<Session> createSession(String databaseName);
```
meltsufin: Need to clarify that this is the fully qualified database name and what the format is.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -40,26 +43,39 @@
...
 
-  private final SpannerStub spanner;
+  private SpannerStub spanner;
 
   /**
-   * Initializes the Cloud Spanner gRPC async stub.
+   * Creates the gRPC stub with authentication.
+   * @throws IOException if credentials could not be acquired
    */
-  public GrpcClient() throws IOException {
```
meltsufin: What's wrong with the constructor potentially throwing an exception?

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -40,26 +43,39 @@
...
     this.spanner = SpannerGrpc.newStub(channel)
         .withCallCredentials(callCredentials);
   }
 
+  /**
+   * Creates a Spanner session to be used for any future operations.
+   * @param databaseName Fully qualified database name in {@code project/instance/database} format.
+   * @return {@link Mono} of the created session.
+   */
+  public Mono<Session> createSession(String databaseName) {
```
meltsufin: `@Override` is missing, and you don't really need to repeat the javadoc if you have it in the interface.

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
meltsufin: Can this method be just turned into its own class like `ObservableMono` with a constructor that takes `Consumer<StreamObserver<ResponseT>> remoteCall`?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
elefeint: Convenience, two-fold:
1) every other method will need to do the same mapping from session to name.
2) testing

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
elefeint: Don't need the getters beyond the fully qualified one, but may be nice to have for logging. What's the harm?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactory.java
elefeint: Will do; I'll have to instantiate the real one `SpannerConnectionFactoryProvider` then.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
elefeint: I would have liked to see the other options have some sort of convention. I can remove the prefix, but then "DRIVER_NAME" and "PROJECT" become semantically conflated, even though one is a key and one is a value.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
elefeint: Personal preference, I guess. I'll put it back.

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
elefeint: That's where I started, but creating an object and immediately calling a single method on it looked weird (and making a gRPC call from a constructor also felt weird).

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
meltsufin: > Convenience, two-fold:
> 
> 1. every other method will need to do the same mapping from session to name.

Can you give an example?

> 2. testing

If the session is passed in through the constructor, then the caller would already know it.

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
meltsufin: What single method would you be calling on it? `ObservableMono` would implement `Mono` itself.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
elefeint: Starting a transaction needs to pass a session name as part of request. So you would do getSessionName().map(name -> /* form the StartTransaction request and call over gRPC here*/);
You can do the same with just the session Mono but really the name is the only useful bit of information a session, might as well make it simple to access.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
ChengyuanZhao: @meltsufin  it looks exactly right in terms of life-cycle of the object, but its features are completely nonexistent !

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactory.java
```diff
@@ -30,14 +32,20 @@
...
+    this.config = Assert.requireNonNull(config, "Spanner configuration must not be null");
   }
 
   @Override
-  public Publisher<? extends Connection> create() {
+  public Publisher<SpannerConnection> create() {
 
-    return Mono.just(new SpannerConnection());
+    return Mono.fromSupplier(() -> {
+      Mono<Session> session = this.client.createSession(config.getFullyQualifiedDatabaseName());
```
meltsufin: You can do this in one line

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
meltsufin: The type of the field informs what it is: `String` vs `Option<String>`. Using `OPTION_PROJECT` to me is as akin to `STRING_DRIVER_NAME`.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
```diff
@@ -34,9 +38,25 @@
...
   @Override
   public ConnectionFactory create(ConnectionFactoryOptions connectionFactoryOptions) {
-    return new SpannerConnectionFactory(null);
+    SpannerConnectionConfiguration config = new SpannerConnectionConfiguration.Builder()
+        .setProjectId(connectionFactoryOptions.getValue(OPTION_PROJECT))
+        .setInstanceName(connectionFactoryOptions.getValue(OPTION_INSTANCE))
+        .setDatabaseName(connectionFactoryOptions.getValue(DATABASE))
+        .build();
+    try {
+      return new SpannerConnectionFactory(new GrpcClient(), config);
```
meltsufin: `GrpcClient` should already be provided to the factory, otherwise there is no way to override the client you're using and the code is untestable.

elefeint: Provided by whom? Connection Factory Providers are loaded by R2DBC [dynamically](https://github.com/r2dbc/r2dbc-spi/blob/4058a9b75bca3f4ce6c926e803c1b5a1d2b14ce4/r2dbc-spi/src/main/java/io/r2dbc/spi/ConnectionFactories.java#L119).

We can abstract this into a provider for testability, but GrpcClient will have to be instantiated in some glue code somewhere.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
```diff
@@ -34,9 +38,25 @@
...
+    } catch (IOException e) {
+      // TODO: log and return null instead?
```
mp911de: No. Always throw an exception if an exception happens.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
```diff
@@ -34,9 +38,25 @@
   /** R2DBC driver name for Google Cloud Spanner. */
   public static final String DRIVER_NAME = "spanner";
```
mp911de: There are two sides to configuration: Programmatic and URL-based. Both ways end up with providing a `ConnectionFactoryOptions` object.

Maybe you should start drawing of a R2DBC URL for Spanner would look like.

How about `r2dbc:spanner:<project>//<instance>/<database>`?

This way, `<project>` gets auto-propagated to `ConnectionFactoryOptions.PROTOCOL`, `<instance>` to `HOST` and `<database>` to `DATABASE`.

You could also have `r2dbc:spanner:///<database>?project=&instance=` but for me that reads sort of strange.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
```diff
@@ -34,9 +38,25 @@
...
+  public static final Option<String> OPTION_PROJECT = Option.valueOf("project");
+
+  /** Option name for GCP Spanner instance. */
+  public static final Option<String> OPTION_INSTANCE = Option.valueOf("instance");
+
   @Override
   public ConnectionFactory create(ConnectionFactoryOptions connectionFactoryOptions) {
-    return new SpannerConnectionFactory(null);
+    SpannerConnectionConfiguration config = new SpannerConnectionConfiguration.Builder()
+        .setProjectId(connectionFactoryOptions.getValue(OPTION_PROJECT))
```
mp911de: That should be `getRequiredValue(…)` to fail if a value is absent.

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
mp911de: 👍

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
```diff
@@ -0,0 +1,57 @@
...
+          sink.success(response);
+        }
+
+        @Override
+        public void onError(Throwable throwable) {
+          sink.error(throwable);
+        }
+
+        @Override
+        public void onCompleted() {
```
mp911de: I'm not familiar with `StreamObserver` whether it can complete without emitting a value. In that case, you'd need to track state whether an element was emitted or not.

pom.xml
elefeint: Added an issue to track removal (#29).

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactory.java
elefeint: This is great. It means that `SpannerConnection` creation is naturally deferred, so now `SpannerConnection` can take an unwrapped Session as a constructor argument.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
elefeint: Sounds great; I created a tracking issue #30.
It may also make sense to do `r2dbc:spanner//project:instance/database`.

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
elefeint: `StreamObserver` is used for both unary (returning single value) and streaming gRPC calls, so in principle, its contract allows it to complete without a value. In practice, onCompleted() will not be invoked for a unary call at all.

I am going to handle the behavior just in case by propagating a Mono error when `onComplete` is called with no value/exception seen.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
elefeint: Created tracking issue (#31).

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
meltsufin: I didn't realize that. Thanks for the clarification.

PR #35, opened 2019-05-09
Implement Basic Transactions

dzou:
Implements basic transaction operations:

`begin`, `commit`, and `close`.

Defaults to read-write transactions only for now.


src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -63,6 +71,33 @@ public GrpcClient() throws IOException {
...
 
+  @Override
+  public Mono<Transaction> beginTransaction(Session session) {
+    BeginTransactionRequest beginTransactionRequest =
+        BeginTransactionRequest.newBuilder()
+            .setSession(session.getName())
+            .setOptions(
+                TransactionOptions
+                    .newBuilder()
+                    .setReadWrite(ReadWrite.getDefaultInstance()))
```
elefeint: Want to add a tracking issue to implement read-only transactions?

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
```diff
@@ -17,30 +17,67 @@
...
   }
+
+  @Test
+  public void beginAndCommitTransactions() {
+    SpannerConnection connection = new SpannerConnection(mockClient, TEST_SESSION);
+
+    Mono.from(connection.commitTransaction()).block();
+    verify(this.mockClient, never()).commitTransaction(any(), any());
+
+    Mono.from(connection.beginTransaction()).block();
```
elefeint: Optional: can try Mono.and() to wait for completion of both at the same time.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -63,6 +71,33 @@ public GrpcClient() throws IOException {
         .withCallCredentials(callCredentials);
   }
 
+  @Override
+  public Mono<Transaction> beginTransaction(Session session) {
+    BeginTransactionRequest beginTransactionRequest =
```
ChengyuanZhao: Is there a `Mono.defer` or something that you could put this request construction in?

elefeint: That's a good point, so that no work would be done until the returned mono is subscribed to. 

We could also keep GrpcClient in imperative-light style, where an action is triggered as soon as methods are called, but then instead defer the invocation of client methods in, say, `Connection`.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -18,69 +18,91 @@
...
     this.client = client;
-
     this.session = session;
+    this.currentTransaction = Mono.empty();
   }
 
+  @Override
   public Publisher<Void> beginTransaction() {
-    return null;
+    currentTransaction = client.beginTransaction(session);
```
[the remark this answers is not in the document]
dzou: This one is a little tricky because Spanner GRPC client does not offer a `getTransactionStatus()` functionality. the client library tracks the status of the transaction based upon whether commit/rollback was successful or if an exception was thrown (in which case the status is "aborted").

I wanted to just keep it simple for now and delegating responsibility of being in a valid transaction state to the user.

I'll create an issue to track as a reminder if we ever want to maintain transaction state for the future.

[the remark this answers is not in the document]
elefeint: Handling it in a followup makes sense to me.
We could track it as "if currentTransaction mono is empty, there is no transaction", and set the currentTransaction to empty after every rollback and commit.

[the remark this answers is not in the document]
dzou: Yeah, I like that idea. Added your comment to the tracking issue #39.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
dzou: Done. #40.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
dzou: Noted. However I wasn't sure if Mono.and(..) waited for the first mono to complete or if both are executed simultaneously. In this case beginTransaction would have to complete first before commit. I suppose Mono.then(..) could be used?

In any case I didn't end up like doing: `Mono.then(Mono.from(...))`, so will keep it like this.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -18,69 +18,91 @@
...
+    return currentTransaction.then();
   }
 
-  public Publisher<Void> close() {
-    return null;
+  @Override
+  public Publisher<Void> commitTransaction() {
+    return currentTransaction
+        .doOnNext(transaction -> client.commitTransaction(session, transaction))
```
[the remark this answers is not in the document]

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -18,69 +18,93 @@
...
+import org.slf4j.LoggerFactory;
+import reactor.core.publisher.Mono;
 
 /**
  * {@link Connection} implementation for Cloud Spanner.
  */
 public class SpannerConnection implements Connection {
 
-  private Session session;
+  private final Logger logger = LoggerFactory.getLogger(this.getClass());
```
meltsufin: You can use Lombok to avoid this line and just add the `@Slf4j` annotation to the class.

dzou:
> Now that the integration test is live, does an integration test change make sense in the same PR or in a followup?

Done. I extended the test to cover session management better. As far as begin/commit transactions, I think we'll only be able to test those once we are able to run some read/write queries.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -106,4 +157,10 @@ public void beforeStart(ClientCallStreamObserver<ExecuteSqlRequest> requestStrea
       this.spanner.executeStreamingSql(request, clientResponseObserver);
     });
   }
+
+  @Override
+  public Mono<Void> close() {
+    return null;
```
meltsufin: I think you need to shutdown the gRPC channel here - `channel.shutdownNow()`

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
dzou: Noted. I ultimately decided to keep the original style since using this requires adding a new library to the pom, and we want to keep the number of dependencies as lightweight as possible.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -18,69 +18,93 @@
...
+        .flatMap(transaction -> client.commitTransaction(session, transaction))
+        .switchIfEmpty(Mono.fromRunnable(() ->
+            logger.warn("commitTransaction() is a no-op; called with no transaction active.")))
+        .then();
   }
 
-  public Publisher<Void> commitTransaction() {
-    return null;
+  @Override
+  public Publisher<Void> close() {
```
elefeint: Defer this, too, so that the session is only deleted on subscribe.

elefeint: Never mind, I got to the defers in the client itself.

PR #41, opened 2019-05-13
Add Rollbacks to Spanner Connection

dzou:
This adds `rollback()` to the SpannerConnection implementation.

Contributes to #35.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
@@ -70,6 +70,15 @@ public class SpannerConnection implements Connection {
   }
 
   @Override
+  public Publisher<Void> rollbackTransaction() {
+    return currentTransaction
+        .flatMap(transaction -> client.rollbackTransaction(session, transaction))
+        .switchIfEmpty(Mono.fromRunnable(() ->
+            logger.warn("rollbackTransaction() is a no-op; called with no transaction active.")))
+        .then();
+  }
+
+  @Override
   public Publisher<Void> close() {
     return client.deleteSession(session);
   }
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -69,6 +69,15 @@ public SpannerConnection(Client client, Session session) {
         .then();
   }
 
+  @Override
+  public Publisher<Void> rollbackTransaction() {
+    return currentTransaction
+        .flatMap(transaction -> client.rollbackTransaction(session, transaction))
```
meltsufin: I may not be understanding Reactor well enough yet, but why `flatMap` vs `map`? There is only one item generated here. Same comment for the method above.

elefeint: Discussed offline that flatMap unwraps nested publishers. In the process, uncovered an interesting testing issue -- if you apply .then() to a nested mono, then there is no way to tell in test whether the inner Mono or the inner Mono's actual object has completed.

dzou: They are similar, but the difference is that map takes a [synchronous mapping function](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#map-java.util.function.Function-) whereas flatmap is [async mapping function](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#flatMap-java.util.function.Function-); i.e. you take something and you map it to another Mono then unwrap the result async.

dzou:
Had to fix checkstyle. PTAL.

PR #42, opened 2019-05-14
Add a test for GrpcClient.executeStreamingSql

### LINKED ISSUE #24: Add a test for GrpcClient.executStreamingSql
Per @mp911de's [comment](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/23#discussion_r281350638):
Query methods are good candidates to be tested via the Reactive Streams TCK (see https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck).

meltsufin:
This test involes starting an embedded gRPC server because there is no
way to mock `SpannerStub`, which a final class.

Fixes #24.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
@@ -157,34 +157,55 @@ public class GrpcClient implements Client {
   @Override
   public Publisher<PartialResultSet> executeStreamingSql(ExecuteSqlRequest request) {
     return Flux.create(sink -> {
-      ClientResponseObserver<ExecuteSqlRequest, PartialResultSet> clientResponseObserver =
-          new ClientResponseObserver<ExecuteSqlRequest, PartialResultSet>() {
-            @Override
-            public void onNext(PartialResultSet value) {
-              sink.next(value);
-            }
+      SinkResponseObserver responseObserver = new SinkResponseObserver<>(sink);
 
-            @Override
-            public void onError(Throwable t) {
-              sink.error(t);
-            }
+      sink.onCancel(
+          () -> responseObserver.getRequestStream().cancel("Flux requested cancel.", null));
 
-            @Override
-            public void onCompleted() {
-              sink.complete();
-            }
+      this.spanner.executeStreamingSql(request, responseObserver);
 
-            @Override
-            public void beforeStart(ClientCallStreamObserver<ExecuteSqlRequest> requestStream) {
-              requestStream.disableAutoInboundFlowControl();
-              sink.onRequest(demand -> requestStream.request((int) demand));
-              sink.onCancel(() -> requestStream.cancel(null, null));
-            }
-
-          };
-      this.spanner.executeStreamingSql(request, clientResponseObserver);
+      // must be invoked after the actual method so that the stream is already started
+      sink.onRequest(demand -> responseObserver.getRequestStream().request((int) demand));
     });
   }
+
+  private static final class SinkResponseObserver<ReqT, RespT> implements
+      ClientResponseObserver<ReqT, RespT> {
+
+    private FluxSink<RespT> sink;
+    private ClientCallStreamObserver<ReqT> requestStream;
+
+    public SinkResponseObserver(FluxSink<RespT> sink) {
+      this.sink = sink;
+    }
+
+    @Override
+    public void onNext(RespT value) {
+      sink.next(value);
+    }
+
+    @Override
+    public void onError(Throwable t) {
+      sink.error(t);
+    }
+
+    @Override
+    public void onCompleted() {
+      sink.complete();
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


meltsufin:
The test currently fails due to some race condition in `beforeStart`.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -79,31 +90,52 @@ public GrpcClient() throws IOException {
...
-              requestStream.disableAutoInboundFlowControl();
-              sink.onRequest(demand -> requestStream.request((int) demand));
-              sink.onCancel(() -> requestStream.cancel(null, null));
-            }
-          };
-      this.spanner.executeStreamingSql(request, clientResponseObserver);
+      SinkResponseObserver responseObserver = new SinkResponseObserver<>(sink);
+
+      sink.onCancel(
+          () -> responseObserver.getRequestStream().cancel("Flux requested cancel.", null));
```
dzou: I think this line should be moved back to the beforeStart() in order to keep the requestStream.cancel() call closer to where the variable is initialized.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -79,31 +90,52 @@ public GrpcClient() throws IOException {
...
+    private FluxSink<RespT> sink;
+    private ClientCallStreamObserver<ReqT> requestStream;
+
+    public SinkResponseObserver(FluxSink<RespT> sink) {
+      this.sink = sink;
+    }
+
+    @Override
+    public void onNext(RespT value) {
+      sink.next(value);
```
elefeint: this.sink (I know, I was going to enable that style check. It's on my todo list... somewhere).

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
elefeint: Right, `SinkResponseObserver`, is already managing the rest of the sink lifecycle, so it belongs there.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -79,31 +90,52 @@ public GrpcClient() throws IOException {
...
-      this.spanner.executeStreamingSql(request, clientResponseObserver);
+      SinkResponseObserver responseObserver = new SinkResponseObserver<>(sink);
+
+      sink.onCancel(
+          () -> responseObserver.getRequestStream().cancel("Flux requested cancel.", null));
+
+      this.spanner.executeStreamingSql(request, responseObserver);
+
+      // must be invoked after the actual method so that the stream is already started
+      sink.onRequest(demand -> responseObserver.getRequestStream().request((int) demand));
```
elefeint: This discrepancy between Reactor operating on `longs` and gRPC operating on `int`s will bite us eventually. Especially when you get into testing unbounded demand (converting max long of unbounded demand into an int would be wrong).

Can you add a follow-up issue to add unbounded demand handling?

elefeint: Also, instead of exposing the implementation detail `getRequestStream()`, perhaps we could add a method `request(long demand)` directly into `SinkResponseObserver`.

src/test/java/com/google/cloud/spanner/r2dbc/client/GrpcClientTest.java
```diff
@@ -0,0 +1,123 @@
...
+    SpannerImplBase spannerSpy = doTest(new SpannerImplBase() {
+          @Override
+          public void executeStreamingSql(ExecuteSqlRequest request,
+              StreamObserver<PartialResultSet> responseObserver) {
+            responseObserver.onNext(PartialResultSet.newBuilder().build());
+            responseObserver.onCompleted();
+          }
+        },
+        // call the method under test
+        grpcClient -> Flux.from(grpcClient.executeStreamingSql(request)).blockFirst()
```
elefeint: Verify result (might have to customize the partial result set for that)?

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -79,31 +90,52 @@ public GrpcClient() throws IOException {
...
     });
   }
+
+  private static final class SinkResponseObserver<ReqT, RespT> implements
+      ClientResponseObserver<ReqT, RespT> {
+
+    private FluxSink<RespT> sink;
```
elefeint: Should I change the ObservableReactiveUtil to also say `ReqT`/`ResT`? I went with more verbose version `RequestT`, but I'd  like to be consistent.

meltsufin: Nah, I'll change mine to match your convention.

PR #45, opened 2019-05-14
Pass Credentials through GrpcClient

### LINKED ISSUE #31: Take credential creation out of GrpcClient
Per Dan's [comment](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/26#discussion_r281736123) on #26.

Credential creation can cause a confounding condition (yes, I am just here for the alliteration. I'll show myself out) because of IOException thrown when searching for credentials on file system. A provider will give us flexibility to swap out credential strategies, and would be a good place to corral the checked exception.

dzou:
Refactor GrpcClient to accept credentials through the constructor.

Fixes #31.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
@@ -16,7 +16,11 @@
 
 package com.google.cloud.spanner.r2dbc;
 
+import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.spanner.r2dbc.util.Assert;
+import java.io.FileInputStream;
+import java.io.IOException;
+import javax.annotation.Nullable;
 
 /**
  * Configurable properties for Cloud Spanner.
@@ -28,15 +32,21 @@ public class SpannerConnectionConfiguration {
 
   private final String fullyQualifiedDbName;
 
+  private final String credentialsLocation;
+
   /**
    * Basic property initializing constructor.
    *
    * @param projectId GCP project that contains the database.
    * @param instanceName instance to connect to
    * @param databaseName database to connect to.
+   * @param credentialsLocation GCP credentials to authenticate service calls with.
    */
   private SpannerConnectionConfiguration(
-      String projectId, String instanceName, String databaseName) {
+      String projectId,
+      String instanceName,
+      String databaseName,
+      @Nullable String credentialsLocation) {
 
     Assert.requireNonNull(projectId, "projectId must not be null");
     Assert.requireNonNull(instanceName, "instanceName must not be null");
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -28,22 +32,37 @@
...
    * @param instanceName instance to connect to
    * @param databaseName database to connect to.
+   * @param credentialsLocation GCP credentials to authenticate service calls with.
    */
   private SpannerConnectionConfiguration(
-      String projectId, String instanceName, String databaseName) {
+      String projectId,
+      String instanceName,
+      String databaseName,
+      @Nullable String credentialsLocation) {
```
meltsufin: If it's optional, either use a setter or overload the constructor.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -28,22 +32,37 @@
...
     Assert.requireNonNull(databaseName, "databaseName must not be null");
 
     this.fullyQualifiedDbName = String.format(
         FULLY_QUALIFIED_DB_NAME_PATTERN, projectId, instanceName, databaseName);
+    this.credentialsLocation = credentialsLocation;
+  }
+
+  GoogleCredentials createCredentials() throws IOException {
+    if (credentialsLocation != null) {
+      return GoogleCredentials.fromStream(new FileInputStream(credentialsLocation));
```
meltsufin: In Spring Cloud GCP, we also support providing the `encodedKey` directly, which is useful when one is storing the credentials somewhere other than a file.
In any case, why not just allow the user to pass in `GoogleCredentials` object, which they can build however they like.

elefeint: I agree. It's best for the driver to not try to infer state from the environment. We can always create a starter/autoconfiguration in `spring-cloud-gcp` to support common usages for the driver.

elefeint:
I agree with injecting credentials instead of figuring them out in the driver.
Since `ConnectionFactoryOptions` can be of any type, the end user can directly pass credentials when creating a connection factory.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
elefeint: this is a private constructor, though. The actual work is done in the builder.
I suppose we can have a private setter, but it makes the object not completely immutable.

meltsufin: fair point!

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
meltsufin: I think using the application default credentials when the user doesn't specify the credentials object is OK. It's a really nice feature that's available across the board on GCP.

elefeint: Hmm, that's true, and that only requires a dependency on google-auth (although the dependency name `google-auth-library-oauth2-http` might be a tad confusing in the context of non-http application. I wonder if there is a leaner dependency we could use.

PR #51, opened 2019-05-15
Implement Spanner Row and Column Metadata

### LINKED ISSUE #34: Row and Rowmetadata implementation

### LINKED ISSUE #50: Extract column names to indices once and store in metadata
https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/46/files#r284388703

SpannerRow is converting column names into column indexes. This work can be done once.
This may require that SpannerRow takes a wrapper object, and not just StructType.

dzou:
This implements the `SpannerRowMetadata` and `SpannerColumnMetadata` objects. Integrates them with `SpannerRow`.

Fixes #34.
Fixes #50.

### DIFF
--- a/src/test/java/com/google/cloud/spanner/r2dbc/SpannerColumnMetadataTest.java
+++ b/src/test/java/com/google/cloud/spanner/r2dbc/SpannerColumnMetadataTest.java
@@ -18,6 +18,7 @@ package com.google.cloud.spanner.r2dbc;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.google.spanner.v1.StructType.Field;
 import org.junit.Test;
 
 /**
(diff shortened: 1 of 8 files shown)


src/test/java/com/google/cloud/spanner/r2dbc/SpannerColumnMetadataTest.java
```diff
@@ -27,8 +28,7 @@
 
   @Test
   public void getNameDummyImplementation() {
```
elefeint: Rename test from "dummy" now that there is real logic.
Also, I Can Has Moar tests (happy path for `getName()`, happy path for getting metadata)? It's tested in `SpannerRowMetadataTest`, but it's nice to have dedicated unit tests.

dzou: Done, added test for `getName()` and `getNativeTypeMetadata()`.

dzou:
> Looks good, just add a few more trivial tests for column.
> 
> _Non-actionable_
> We might eventually also need to include [ResultSetStats](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#resultsetstats) to support update queries. Or maybe the not, since the query vs update result set information is already encoded by way of picking the right `SpannerResult` constructor.

Noted. Regarding `ResultSetStats` it might be possible that we don't need it given what you said about how the row count is encoded in the constructor. We'll keep the door open on this and add it in if we determine its necessary for some interface method.

PR #54, opened 2019-05-16
Assemble rows from partial results

### LINKED ISSUE #33: Client assembles `List<Value>` from partial result sets implementation

ChengyuanZhao:
assembles partial result sets into rows and unit test
fixes #33


src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
```diff
@@ -72,6 +83,16 @@
...
   Publisher<PartialResultSet> executeStreamingSql(ExecuteSqlRequest request);
 
+  /**
+   * Execute a streaming query and get discrete rows as lists of protobuf values.
+   *
+   * @param partialResultSetPublisher a publisher of partial result sets.
+   * @return a publisher of individual rows where each row is a list of column protobuf values and
+   *        the metadata of the columns in the result.
+   */
+  Tuple2<Mono<ResultSetMetadata>, Flux<List<Value>>> assembleRowsFromPartialResults(
```
dzou: Should this method just be merged into `executeStreamingSql`? In practice I don't see anyone calling `executeStreamingSql` by itself because one is unable to reason about the results without partial result reassembly.

dzou: I think the return type instead of `Tuple2` can just be `SpannerResult` since those are the two inputs needed for SpannerResult class.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -173,8 +192,148 @@ public void beforeStart(ClientCallStreamObserver<ExecuteSqlRequest> requestStrea
     });
   }
 
+  @Override
+  public Tuple2<Mono<ResultSetMetadata>, Flux<List<Value>>> assembleRowsFromPartialResults(
+      Publisher<PartialResultSet> partialResultSetPublisher) {
```
dzou: Since the Client is going to be used internally by us (and not exposed to users), I think it's fine to keep the input parameter of type `Flux<PartialResultSet>`; save yourself conversion from Publisher to flux.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -173,8 +192,148 @@ public void beforeStart(ClientCallStreamObserver<ExecuteSqlRequest> requestStrea
...
+        .map(PartialResultSet::getMetadata);
+    return Tuples.of(resultSetMetadataMono,
+        streamingListValueRows(partialResultSetFlux, resultSetMetadataMono));
+  }
+
+  private Flux<List<Value>> streamingListValueRows(Flux<PartialResultSet> partialResultSetFlux,
+      Mono<ResultSetMetadata> resultSetMetadataMono) {
+
+    AtomicReference<
+        Flux<Tuple2<ResultSetMetadata, PartialResultSet>>> zipped =
```
dzou: could this var be moved inside the `Flux.create` to avoid declaring it an Atomic reference?

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -173,8 +192,148 @@ public void beforeStart(ClientCallStreamObserver<ExecuteSqlRequest> requestStrea
...
+        new AtomicReference<>(
+            Flux.combineLatest(resultSetMetadataMono, partialResultSetFlux, Tuples::of));
+
+    return Flux.create(sink -> {
+      AtomicBoolean prevIsChunk = new AtomicBoolean(false);
```
dzou: same with `prevIsChunk`, `incompletePiece`, and `incompletePieceKind`; you might be able to move this inside the zipped.set(...)'s lambda and not have to declare them as atomic since it doesn't appear to be needed outside that scope.

src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
ChengyuanZhao: Good point yea. originally i thought maybe the Statement or Connection would do that, but it's simple enough here too.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
ChengyuanZhao: yes!

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
ChengyuanZhao: They're used in the parameter of zipped.doOnNext and they must be around for longer than one invocation of that anonymous function...

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -173,8 +192,144 @@ public void beforeStart(ClientCallStreamObserver<ExecuteSqlRequest> requestStrea
...
+                PartialResultSet partialResultSet = t.getT2();
+                int availableCount = partialResultSet.getValuesCount();
+
+                if (prevIsChunk.get()) {
+                  Value firstPiece = partialResultSet.getValues(0);
+
+                  // Concat code from client lib
+                  if (incompletePieceKind.get() == KindCase.STRING_VALUE) {
+                    incompletePiece.set(incompletePiece.get() + firstPiece.getStringValue());
+                  } else {
```
elefeint: What about object concatenation? Docs say [object splitting is supported](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.PartialResultSet), but I am not sure what an "object" value sent from Spanner even looks like from Java side.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -173,8 +192,144 @@ public void beforeStart(ClientCallStreamObserver<ExecuteSqlRequest> requestStrea
...
+                /* Only the final value can be chunked, and only the first value can be
+                  a part of a previous chunk, so the pieces in the middle are always
+                  whole values.
+                * */
+                for (int i = 1; i < availableCount - 1; i++) {
+                  appendToRow.accept(partialResultSet.getValues(i));
+                }
+
+                // this final piece is the start of a new incomplete value
+                if (!prevIsChunk.get() && partialResultSet.getChunkedValue()) {
```
elefeint: What if previous was also a chunk? (in other words, a looooong value is split over 3 different PartialResultSets, with the middle PartialResultSet only containing 1 value)?
Would likely need to append to incomplete piece.

ChengyuanZhao: Yes this is supported. One of the tests i have has "abc" broken as "a" , "b" , and "c" in 2 PRs.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
ChengyuanZhao: TBH I don't think this is actually supported in code. 

Looking at the client lib they support merging of 2 things 
```
  private boolean isMergeable(KindCase kind) {
    return kind == KindCase.STRING_VALUE || kind == KindCase.LIST_VALUE;
  }
```

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -173,8 +192,144 @@ public void beforeStart(ClientCallStreamObserver<ExecuteSqlRequest> requestStrea
...
+      Consumer<Value> appendToRow = val -> {
+        currentRow.get().add(val);
+        if (currentRow.get().size() == rowSize.get()) {
+          sink.next(currentRow.get());
+          currentRow.set(new ArrayList<>());
+        }
+      };
+
+      Flux<Tuple2<ResultSetMetadata, PartialResultSet>> zipped =
+          Flux.combineLatest(resultSetMetadataMono, partialResultSetFlux, Tuples::of)
```
meltsufin: You probably don't need to split metadata and the rest of result set into a tuple, and instead just keep track of metadata after you see it here.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -173,8 +192,144 @@ public void beforeStart(ClientCallStreamObserver<ExecuteSqlRequest> requestStrea
...
+              .doOnNext(t -> {
+                if (rowSize.get() == -1) {
+                  rowSize.set(t.getT1().getRowType().getFieldsCount());
+                }
+                PartialResultSet partialResultSet = t.getT2();
+                int availableCount = partialResultSet.getValuesCount();
+
+                if (prevIsChunk.get()) {
```
meltsufin: Try to break up this method into smaller ones.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -173,8 +192,144 @@ public void beforeStart(ClientCallStreamObserver<ExecuteSqlRequest> requestStrea
     });
   }
 
+  @Override
+  public Tuple2<Mono<ResultSetMetadata>, Flux<List<Value>>> assembleRowsFromPartialResults(
```
meltsufin: make this method private, and create a similar method that doesn't require passing the partial results.

ChengyuanZhao:
@elefeint @dzou Merged !

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -31,61 +35,126 @@
...
+            firstPiece.getListValue().getValuesList());
+      }
     }
 
+    /* if there are more values then it means the incomplete piece is complete.
+    Also, if this PR isn't chunked then it is also complete. */
+    if (availableCount > 1 || !partialResultSet.getChunkedValue()) {
+      if (prevIsChunk) {
+        appendToRow.accept(
+            incompletePieceKind == KindCase.STRING_VALUE
```
elefeint: Can the assembly of the single, previously chunked value go into a helper function, so that the main reassembly logic is easier to follow (the Spanner reassembly rules are complex enough without worrying about the types. Plus, if we ever want to support object reassembly, it would be good to have the type-specific logic encapsulated in a separate method).

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -31,61 +35,126 @@
...
+              + "metadata but it was null."));
+      rowSize = partialResultSet.getMetadata().getRowType().getFieldsCount();
+    }
+    int availableCount = partialResultSet.getValuesCount();
+
+    if (prevIsChunk) {
+      Value firstPiece = partialResultSet.getValues(0);
+
+      // Concat code from client lib
+      if (incompletePieceKind == KindCase.STRING_VALUE) {
```
elefeint: Could this if/else clause go into a helper method `extendChunkedValue()` (come up with a better name)

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -31,61 +35,126 @@
...
-
-  private Value incompleteField;
-
+  int rowSize;
+  boolean prevIsChunk;
+  List<Value> currentRow = new ArrayList<>();
+  Object incompletePiece;
+  KindCase incompletePieceKind;
+
+  BiConsumer<Value, FluxSink<SpannerRow>> appendToRow = (val, sink) -> {
```
elefeint: Now that this logic is in a standalone class, the method reference can become just a member helper method `appendToRow()`.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -31,61 +35,126 @@
...
+    }
 
-      startIndex += this.numFieldsPerRow;
-      endIndex += this.numFieldsPerRow;
+    Value lastVal = partialResultSet.getValues(availableCount - 1);
 
+    // this final piece is the start of a new incomplete value
+    if (!prevIsChunk && partialResultSet.getChunkedValue()) {
+      incompletePieceKind = lastVal.getKindCase();
+      incompletePiece = lastVal.getKindCase() == KindCase.STRING_VALUE ? lastVal.getStringValue() :
```
elefeint: I know this is trivial logic, but could this type-specific bit also go into a helper method, something like `createNewChunkedValue()`?
I am starting to think that we should create a strategy class with three methods  -- `create`, `append` and `complete`, and then have two implementations, one for String and one for List.
Then type-specific functionality can be taken out of this class completely.

ChengyuanZhao: refactored

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
ChengyuanZhao: fixed

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
ChengyuanZhao: fixed

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
ChengyuanZhao: fixed

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -31,61 +35,126 @@
...
-    if (partialResultSet.hasMetadata()) {
-      metadata = new SpannerRowMetadata(partialResultSet.getMetadata());
-      StructType rowType = partialResultSet.getMetadata().getRowType();
-      this.numFieldsPerRow = rowType.getFieldsCount();
-    }
-    // TODO: handle the case where metdata is not available yet
+  public void emitRows(PartialResultSet partialResultSet, FluxSink<SpannerRow> sink) {
     if (metadata == null) {
-      throw new RuntimeException("Metadata failed to arrive with the first PartialResultSet");
+      metadata = new SpannerRowMetadata(Assert.requireNonNull(partialResultSet.getMetadata(),
```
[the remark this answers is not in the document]
ChengyuanZhao: fixed

src/main/java/com/google/cloud/spanner/r2dbc/SpannerRowMetadata.java
```diff
@@ -37,7 +38,7 @@
    */
   private final HashMap<String, Integer> columnNameIndex;
 
-  private ResultSetMetadata rowMetadata;
+  private final ResultSetMetadata resultSetMetadata;
```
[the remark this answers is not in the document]
ChengyuanZhao: fixed

src/main/java/com/google/cloud/spanner/r2dbc/SpannerRow.java
```diff
@@ -59,12 +59,11 @@ public SpannerRow(List<Value> values, SpannerRowMetadata rowMetadata) {
   }
 
   @VisibleForTesting
```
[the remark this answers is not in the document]
ChengyuanZhao: fixed

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -74,6 +75,12 @@ public GrpcClient(GoogleCredentials credentials) {
         .withCallCredentials(callCredentials);
   }
 
+  @VisibleForTesting
+  GrpcClient(ManagedChannel channel, SpannerStub spanner) {
```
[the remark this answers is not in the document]
ChengyuanZhao: removed

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultFluxConverter.java
```diff
@@ -48,41 +48,33 @@ public PartialResultFluxConverter(Flux<PartialResultSet> results) {
...
+      public void onSubscribe(Subscription subscription) {
+        PartialResultFluxConverter.this.spannerSubscription = subscription;
+        // initial result
+        PartialResultFluxConverter.this.spannerSubscription.request(1);
+      }
+
+      @Override
+      public void onNext(PartialResultSet partialResultSet) {
+
+        PartialResultFluxConverter.this.rowExtractor.emitRows(partialResultSet, sink);
```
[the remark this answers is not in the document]
ChengyuanZhao: understood, fixed.!

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -31,61 +32,145 @@
 
   // TODO: this should also track the latest resume_token and return it upon request
 
-
   // this probably does not even need an atomic reference. Double check gRPC java listener
```
dzou: Is this comment still relevant?

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -31,61 +32,145 @@
...
   // implementation, but it should be accessed by a single thread.
   private SpannerRowMetadata metadata = null;
-
-  private int numFieldsPerRow;
-
-  private List<Value> incompleteRow;
-
-  private Value incompleteField;
-
+  int rowSize;
```
dzou: make fields private?

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -31,61 +32,145 @@
...
+      } else {
+        concatLists((List<Value>) incompletePiece,
+            firstPiece.getListValue().getValuesList());
+      }
     }
+  }
 
-    // TODO: store partial row + last field, if chunked.
+  private void setMetadata(PartialResultSet partialResultSet) {
+    if (metadata == null) {
```
dzou: Replace `metadata == null` with `if (partialResultSet.hasMetadata()) {...}`?

You could then omit null check and the `IllegalStateException` check.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultFluxConverter.java
```diff
@@ -48,41 +48,34 @@ public PartialResultFluxConverter(Flux<PartialResultSet> results) {
...
-        }
-      });
-
-
-    });
+    return Flux.create(sink -> results.subscribe(new CoreSubscriber<PartialResultSet>() {
+
+      @Override
+      public void onSubscribe(Subscription subscription) {
+        PartialResultFluxConverter.this.spannerSubscription = subscription;
```
dzou: If u want, you can omit `PartialResultFluxConvert.this.` to reference class variables of the parent from the anonymous class.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
ChengyuanZhao: I'm not sure I follow my dude. We need the metadata for the very first PR or we must throw the exception.

dzou: Got it.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultFluxConverter.java
```diff
@@ -48,41 +48,34 @@ public PartialResultFluxConverter(Flux<PartialResultSet> results) {
...
+      public void onSubscribe(Subscription subscription) {
+        spannerSubscription = subscription;
+        // initial result
+        spannerSubscription.request(1);
+      }
+
+      @Override
+      public void onNext(PartialResultSet partialResultSet) {
+
+        PartialResultFluxConverter.this.rowExtractor.emitRows(partialResultSet).stream()
```
meltsufin: @elefeint If it's a 1-to-1 relationship between PartialResultFluxConverter and PartialResultRowExtractor, can they just be merged into one class?

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -29,63 +30,143 @@
  */
 public class PartialResultRowExtractor {
```
meltsufin: @elefeint Is there an intent to have multiple implementations of this class? In that case you probably need an interface.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -29,63 +30,143 @@
...
-    if (partialResultSet.hasMetadata()) {
-      metadata = new SpannerRowMetadata(partialResultSet.getMetadata());
-      StructType rowType = partialResultSet.getMetadata().getRowType();
-      this.numFieldsPerRow = rowType.getFieldsCount();
+  public List<SpannerRow> emitRows(PartialResultSet partialResultSet) {
+    List<SpannerRow> sink = new ArrayList<>();
+    setMetadata(partialResultSet);
+    int availableCount = partialResultSet.getValuesCount();
+
+    concatFirstIncompletePiece(partialResultSet);
```
meltsufin: Maybe take the condition out of the method to aid readability.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -29,63 +30,143 @@
...
+
+    /* if there are more values then it means the incomplete piece is complete.
+    Also, if this PR isn't chunked then it is also complete. */
+    if (availableCount > 1 || !partialResultSet.getChunkedValue()) {
+      if (prevIsChunk) {
```
meltsufin: Put this block in a method called `completeValue`.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -29,63 +30,143 @@
...
+        appendToRow(
```
meltsufin: move `appendToRow` below the if-else.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -29,63 +30,143 @@
...
-
-    // TODO: handle partials left over at the end of previous row
-    // TODO: account for chunked values (field split between partial result sets).
-    List<Value> values = partialResultSet.getValuesList();
+    emitMiddleWholePieces(partialResultSet, sink, availableCount);
 
-    int startIndex = 0;
-    int endIndex = this.numFieldsPerRow;
+    Value lastVal = partialResultSet.getValues(availableCount - 1);
+    beginIncompletePiece(partialResultSet, sink, availableCount, lastVal);
```
meltsufin: take condition out here too.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultFluxConverter.java
elefeint: They could, but I would prefer for them to stay as two classes because they handle separate concerns. I favor the single responsibility principle -- if we need to change backpressure logic, it should not cause a change in any parsing logic, and vice versa.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
elefeint: I don't believe that we'll need multiple implementations of the actual parsing logic. We could certainly introduce an interface to make testing more flexible, though (I think you mentioned this as a possibility last week).

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultFluxConverter.java
meltsufin: That sounds reasonable but the issue is that the two classes are still tightly coupled because there is no interface for the `PartialResultRowExtractor`, and `PartialResultFluxConverter` actually directly instantiates it, instead of letting the caller decide which extractor to use.
```
private PartialResultRowExtractor rowExtractor = new PartialResultRowExtractor();
```

elefeint: Good point, then let's inject it through constructor?

elefeint: As an interface.

meltsufin: Looking at this more, I actually think it might make sense to make `PartialResultFluxConverter` implement `CoreSubscriber`. If you look at it, it currently doesn't almost nothing besides creating the anonymous `CoreSubscriber` implementation that does the conversion. The only thing it really does is `Flux.create(sink -> results.subscribe(...`, which I think can move to the caller.

ChengyuanZhao: Done

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultFluxConverter.java
```diff
@@ -20,69 +20,50 @@
...
+  @Override
+  public void onSubscribe(Subscription s) {
+    spannerSubscription = s;
+    // initial result
+    spannerSubscription.request(1);
+  }
 
+  @Override
+  public void onNext(PartialResultSet partialResultSet) {
+    PartialResultFluxConverter.this.rowExtractor.emitRows(partialResultSet).stream()
```
elefeint: Don't need `.stream()`; `List` has `forEach()`.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -29,63 +30,141 @@
...
   /**
    * Assembles as many complete rows as possible, given previous incomplete fields and a new
    * {@link PartialResultSet}.
    * @param partialResultSet a not yet processed result set
-   * @return an ordered list of full rows, each containing the row metadata
    */
-  public List<SpannerRow> extractCompleteRows(PartialResultSet partialResultSet) {
-    List<SpannerRow> fullRows = new ArrayList<>();
+  public List<SpannerRow> emitRows(PartialResultSet partialResultSet) {
+    List<SpannerRow> sink = new ArrayList<>();
```
elefeint: Could you rename from `sink` to `rows`?

ChengyuanZhao: done!

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultFluxConverter.java
ChengyuanZhao: done!

ChengyuanZhao:
@meltsufin  any final comments or changes ?

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultFluxConverter.java
```diff
@@ -20,69 +20,49 @@
...
 import org.reactivestreams.Subscription;
 import reactor.core.CoreSubscriber;
-import reactor.core.publisher.Flux;
+import reactor.core.publisher.FluxSink;
 
 /**
  * Provides a stream of Cloud Spanner rows based on the input stream of {@link PartialResultSet}s.
  */
-public class PartialResultFluxConverter {
+public class PartialResultFluxConverter implements CoreSubscriber<PartialResultSet> {
```
meltsufin: I think we can make this class more generic/re-usable as well by making it implement `CoreSubscriber<T>`, but I'll leave to a future PR.

ChengyuanZhao: I can make an issue for that but I wanna ask more about what you mean here:

The reconstruction logic is determined by the rules of the PartialResultSet class. All the rules about how the metadata must come attached to the first PR, to count off rows by number of type objects in the metadata, that each PR has a series of pieces, middle pieces are whole, final pieces can be chunked, etc.. this all is part of the PR specification and doesn't seem generalized to any type T.

PR #56, opened 2019-05-17
Execute statement query + structure for reassembling results

elefeint:
This PR:
* Implements statement execution, so happy path is now testable against real Spanner end-to-end (connection to statement to execution to result extraction).
* Separates  `PartialResultSet`-to-`Row` Flux conversion concerns from record reassembly concerns. 
* Intentionally stays away from thread-safe complications. I need to do a bit more research, but I think it will only be accessed by one thread at a time, in order, given that:
  1.  gRPC [guarantees message ordering within an individual RPC call](https://grpc.io/docs/guides/concepts/#service-definition)
  2. Flux.create() uses a SerializedSink, which serializes calls to onNext().
* That, in turn, means that result set metadata extraction can be done with a simple check for metadata arrival as soon as row conversion logic is invoked (because if it did not show up together with the first row, that's a violation of spanner contract).
* Use `SpannerRow` instead of `List<Value>` to represent row data.
  * the abstraction is easier to reason about
  * allows metadata to be attached to each row early on

What's not in this PR?
* Proper row reassembling logic, since it's analyzed in #54 .
* Handling of update queries.

Changes expected to this PR:
* The changes to GrpcClient/ObservableReactiveUtil will go away once #42 is merged.
* I'll add tests for new functionality.


src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -60,6 +85,16 @@ public Statement bindNull(int i, Class<?> type) {
 
   @Override
   public Publisher<? extends Result> execute() {
-    return Mono.just(new SpannerResult(Mono.fromSupplier(() -> 0)));
+    Flux<PartialResultSet> result
+        = client.executeStreamingSql(this.session, this.transaction, this.sql);
+
+
+    PartialResultFluxConverter rsTracker = new PartialResultFluxConverter(result);
```
elefeint: client will handle the conversion -- move this line to the client (client API will change to expose full rows, and now partial result sets)

src/main/java/com/google/cloud/spanner/r2dbc/SpannerRowMetadata.java
```diff
@@ -37,12 +37,17 @@
    */
   private final HashMap<String, Integer> columnNameIndex;
 
+  private ResultSetMetadata rowMetadata;
```
elefeint: misnamed, and see if it's necessary.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.protobuf.Value;
+import com.google.spanner.v1.PartialResultSet;
+import com.google.spanner.v1.StructType;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * NOT thread-safe. But it likely does not need to be.
+ */
+public class PartialResultRowExtractor {
```
elefeint: create testable interface

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -150,24 +151,35 @@ public GrpcClient(SpannerStub spanner) throws IOException {
...
+    return transaction
+        .map(t -> TransactionSelector.newBuilder().setId(t.getId()).build())
+        .defaultIfEmpty(readOnlySingleUseTransaction())
+        .map(t ->  ExecuteSqlRequest.newBuilder()
+            .setSql(sql)
+            .setSession(session.getName())
+            .setTransaction(t)
+            .build())
+        .flatMapMany(request -> Flux.create(
+            sink -> {
```
elefeint: @meltsufin How do you feel about this flux creation and `SinkResponseObserver` nested class moving into `ObservableReactiveUtil`? 
Because transaction is a mono, we need a pipeline that ends with getting a Flux out of the `ClientResponseObserver` implementation, but with a nested sink in line it looks like a lot of code in a class that's mostly about interacting with low-level Spanner APIs.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerRowMetadata.java
elefeint: it was not.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
meltsufin: Sure, we can move it there, if it will be used in more than one place.

elefeint: It won't... We only have one streaming call.

Issue #59, opened 2019-05-20
Implement retry logic for streamed calls


dzou:
Retry logic for streamed calls might be slightly different than unary calls since streamed calls can fail in the middle of emitted elements; thus previously emitted elements must be carefully discarded. This might require some special treatment.

related to #53

dzou:
Obsolete after #79.

PR #60, opened 2019-05-20
Implement Retries for Unary Calls

### LINKED ISSUE #53: Implement Retry Logic for Failed RPCs
Implement Retry logic for failed RPCs.

This is the case where an RPC might fail due to network errors or other external factors. This excludes the case where a `commit()` fails and is aborted.

See this for reference: https://github.com/googleapis/google-cloud-java/blob/7d47fce81df96e1a0e0270cfe264fcecbfb0bb22/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L165

dzou:
This implements retry logic for unary calls. Fixes #53.

Note: Holding off implementing this for streamed calls for now. Why? Because retrying on Fluxes [resubscribes to the flux](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#retry--); this means that previously emitted elements should be discarded. This case might be slightly different than unary calls. Created #59 to track.


src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
```diff
@@ -0,0 +1,81 @@
...
+      if (statusRuntimeException.getStatus().getCode() == Status.Code.INTERNAL) {
+        if (cause.getMessage().contains("HTTP/2 error code: INTERNAL_ERROR")) {
+          return true;
+        }
+        if (cause.getMessage().contains("Connection closed with unknown cause")) {
+          return true;
+        }
+        if (cause
+            .getMessage()
+            .contains("Received unexpected EOS on DATA frame from server")) {
```
ChengyuanZhao: Could a link be added in the comments or javadoc for this method about where these specific 4 exception descriptions came from?

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
```diff
@@ -0,0 +1,81 @@
...
+          return true;
+        }
+      }
+
+      if (statusRuntimeException.getStatus().getCode() == Code.RESOURCE_EXHAUSTED
+          && extractRetryDelay(statusRuntimeException).toMillis() > 0L) {
```
ChengyuanZhao: Is this delay the recommended retry time? If so, would it make sense to use this as the retry delay instead of using the strategy defined in the other class?

src/test/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtilTest.java
```diff
@@ -56,4 +59,42 @@ public void unaryCallThrowsExceptionIfCompletedWithNoValue() {
         .hasMessage("Unary gRPC call completed without yielding a value or an error");
   }
 
+  @Test
+  public void unaryCallRetries() {
+    StatusRuntimeException retryableError = new StatusRuntimeException(
+        Status.INTERNAL.withDescription("HTTP/2 error code: INTERNAL_ERROR"), null);
+
+    AtomicInteger timesRetried = new AtomicInteger(0);
```
ChengyuanZhao: These "Atomic" things are useful when working w/ anon functions right ?

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
dzou: Yes, I wanted to do it this way initially, but the process of parsing the exception and setting that as the delay is complex; would require code [that looked like](https://projectreactor.io/docs/core/release/reference/#faq.exponentialBackoff):

```
Flux<String> flux =
Flux.<String>error(new IllegalArgumentException())
    .retryWhen(companion -> companion
        .doOnNext(s -> System.out.println(s + " at " + LocalTime.now())) 
        .zipWith(Flux.range(1, 4), (error, index) -> { 
          if (index < 4) return index;
          else throw Exceptions.propagate(error);
        })
        .flatMap(index -> Mono.delay(Duration.ofMillis(index * 100))) 
        .doOnNext(s -> System.out.println("retried at " + LocalTime.now())) 
    );
```

Additionally, it turns out that the retry delay provided in the exception is more of a "guideline" than a special number. The Go client library team elected to implement it client-side for simplicity as well: https://github.com/googleapis/google-cloud-go/issues/1309.

So I ultimately decided with the current approach because it is easier code to maintain; there wasn't enough "win" to be gained from using the retry time in the exception. And don't know if it is a reliable number (i.e. if you must also pass in your retry attempt count, etc.)

ChengyuanZhao: Would something like this work ?

```
  private static Retry getRetry(){
    AtomicReference<Duration> wait = new AtomicReference<>();
    
    Retry r = Retry.onlyIf(retryContext -> {
      wait.set(extractRetryDelay(retryContext.exception()));
      return isRetryable(retryContext.exception());
    });
    
    return (wait.get().toMillis() > 0L ? r.backoff(Backoff.fixed(wait.get())) : r.exponentialBackoffWithJitter(Duration.ofSeconds(1), Duration.ofSeconds(32))).timeout(Duration.ofSeconds(60));
  }
```

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
dzou: Done.

src/test/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtilTest.java
dzou: Indeedliomo

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
ChengyuanZhao: I see. I wonder why we had to throw out everything from the client lib (like their implementation of isRetryable and exception cause matchers) even though all we wanted to do was use some gapic methods.

dzou:
> I think this looks good. The exception checking matches the client lib and is great.
> 
> The retry settings from the gapic yaml file says a multiplier of 1.3.. does that mean they multiply the wait interval by that each failure? That would result in many more attempts in 60 seconds than a typical exponential backoff would, right ?

Ah, so I've chosen a timeout of 60 seconds where gapic actually does 600 seconds (10 minutes). What are your thoughts; should I set the time out to 600 seconds? Just seemed like overkill.

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
```diff
@@ -16,20 +16,31 @@
...
  * Converter from a gRPC async calls to Reactor primitives ({@link Mono}).
  */
 public class ObservableReactiveUtil {
 
+  // Retry settings inspired from:
+  // https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/spanner_gapic.yaml#L48
+  private static final Retry retryStrategy =
+      Retry.onlyIf(retryContext -> isRetryable(retryContext.exception()))
+          .exponentialBackoffWithJitter(Duration.ofSeconds(1), Duration.ofSeconds(32))
+          .timeout(Duration.ofSeconds(60));
```
elefeint: Created #63 to track propagating these settings from connection options.

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
```diff
@@ -0,0 +1,84 @@
...
+   *
+   * <p>Derived from google-cloud-java/SpannerExceptionFactory.java:
+   * https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
+   */
+  static boolean isRetryable(Throwable cause) {
+    if (cause instanceof StatusRuntimeException) {
+      StatusRuntimeException statusRuntimeException = (StatusRuntimeException) cause;
+
+      if (statusRuntimeException.getStatus().getCode() == Status.Code.INTERNAL) {
+        if (cause.getMessage().contains("HTTP/2 error code: INTERNAL_ERROR")) {
```
elefeint: Could you stick all valid retryable message fragments into a collection and iterate over it?

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
```diff
@@ -0,0 +1,84 @@
...
+    if (trailers != null && trailers.containsKey(KEY_RETRY_INFO)) {
+      RetryInfo retryInfo = trailers.get(KEY_RETRY_INFO);
+      if (retryInfo.hasRetryDelay()) {
+        com.google.protobuf.Duration protobufDuration = retryInfo.getRetryDelay();
+        return Duration.ofSeconds(protobufDuration.getSeconds())
+            .withNanos(protobufDuration.getNanos());
+      }
+    }
+
+    return Duration.ofSeconds(-1);
```
elefeint: Why not return null to signal the absence of delay?

src/test/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtilTest.java
```diff
@@ -56,4 +59,42 @@ public void unaryCallThrowsExceptionIfCompletedWithNoValue() {
...
+          if (timesRetried.get() < 3) {
+            obs.onError(retryableError);
+            timesRetried.addAndGet(1);
+          } else {
+            obs.onNext(100);
+          }
+        });
+
+    int result = resultMono.block();
+    assertThat(result).isEqualTo(100);
```
elefeint: assert that `timesRetried` is exactly 3? That is, it did not continue retrying.

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
dzou: Done.

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
dzou: Done.

src/test/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtilTest.java
dzou: Done.

PR #61, opened 2019-05-21
Refactor PartialResultFluxConverter to a generic ConvertingFluxAdapter

meltsufin:


src/main/java/com/google/cloud/spanner/r2dbc/util/OneToManyFluxMapper.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private Subscription subscription;
+
+  /**
+   * Constructor.
+   *
+   * @param sink the flux sink to which converted object will be forwarded
+   * @param converter the converter that will map input objects to lists of output objects
+   */
+  public OneToManyFluxMapper(FluxSink<R> sink, Function<T, List<R>> converter) {
```
elefeint: It's more many-to-many (one Spanner row could arrive in many `PartialResultSet`s, and one `PartialResultSet` could contain many rows). `converter.apply(input)` returns an empty list if a full row has not been assembled yet. . 
Maybe name it just `FluxMapper` if you choose to keep it generic, or `SpannerRowFluxMapper` if you turn it into a PartialResultSet-specific class.

Also, I wonder if `Adapter` would reflect the purpose better than either `Converter` or `Mapper`, both of which imply 1-1 correspondence.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
```diff
@@ -24,11 +24,12 @@
 import com.google.spanner.v1.PartialResultSet;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Function;
 
 /**
  * NOT thread-safe. But it likely does not need to be.
  */
-public class PartialResultRowExtractor {
+public class PartialResultRowExtractor implements Function<PartialResultSet, List<SpannerRow>> {
```
elefeint: I wonder if we should rename it to imply the side effect. Is there a non-verbose way to say `AccumulatingPartialResultRowProducer`?

src/main/java/com/google/cloud/spanner/r2dbc/util/OneToManyFluxMapper.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.function.Function;
+import org.reactivestreams.Subscription;
+import reactor.core.CoreSubscriber;
+import reactor.core.publisher.FluxSink;
+
+/**
+ * A {@link org.reactivestreams.Subscriber} that given a function {@code Function<T, List<R>>
+ * converter}, maps and forwards a stream  of {@code T} objects to a sink of {@code R} objects.
+ */
+public class OneToManyFluxMapper<T, R> implements CoreSubscriber<T> {
```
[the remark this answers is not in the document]
meltsufin: Even if we never use it for anything else, the advantage is loose coupling. Plus it is much better respects the single responsibility principle which was the original goal of the class anyway.
Going through this exercise also made me wonder if there is something like this already in the Reactor project that we're just missing.

src/main/java/com/google/cloud/spanner/r2dbc/util/OneToManyFluxMapper.java
meltsufin: I was kind of wondering the same thing. The reason I chose to go with one-to-many was because the extractor interface is one in, list out.

src/main/java/com/google/cloud/spanner/r2dbc/util/OneToManyFluxMapper.java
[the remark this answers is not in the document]
elefeint: pubsub maybe.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
meltsufin: Sure, or we can have good javadoc explaining what it does to compensate for the confusing name.

meltsufin:
@elefeint Let me know what you'd like to change. I'm flexible with the naming.

src/main/java/com/google/cloud/spanner/r2dbc/result/PartialResultRowExtractor.java
elefeint: +1 on Javadoc; and remove my "NOT thread-safe..." comment. A better javadoc is my tech debt, anyway, so I can add it in a followup if you'd rather submit this.

src/main/java/com/google/cloud/spanner/r2dbc/util/OneToManyFluxMapper.java
elefeint: How about `ConvertingFluxAdapter`? or just `FluxAdapter`.

PR #62, opened 2019-05-21
Dml support and integration tests! 

### LINKED ISSUE #57: Execute DML queries and handle their return types in Client
So far we have only handled assembling rows from the results of a read query. We will need to decide on how to package the results for a DML query.

ChengyuanZhao:
DML support and tests
fixes #57


src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -225,18 +226,4 @@ public void beforeStart(ClientCallStreamObserver<ReqT> requestStream) {
       }
     });
   }
-
-  /**
-   * Creates a temporary read-only transaction with strong concurrency, which is also the default
-   * for {@code ExecuteStreamingSql} when the transaction field is empty.
-   */
-  private TransactionSelector readOnlySingleUseTransaction() {
```
dzou: Should this be kept here? Because it is possible for statements to be executed outside of a transaction; these would rely on the single-use option.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -93,6 +94,7 @@ public Statement bindNull(int i, Class<?> type) {
     // only a single constructor that takes a flux of rows and a mono of the # rows updated).
     return Mono
         .just(new SpannerResult(
-            Flux.create(sink -> result.subscribe(new PartialResultFluxConverter(sink)))));
+            Flux.create(sink -> result.subscribe(new PartialResultFluxConverter(sink))),
```
dzou: Does result need to be cached here? i.e. if user calls `map` twice, I think the query will get run twice for the same reason that `beginTransaction` was run twice.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -93,6 +94,7 @@ public Statement bindNull(int i, Class<?> type) {
...
+            Mono.create(sink -> result.subscribe(new PartialResultSetStatsConverter(sink)))));
```
dzou: Would the following code be able to replace introducing a new `PartialResultSetStatsConverter` object?

`result.next().flatMap(partialResultSet -> // code that gets the rows updated here)`

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
ChengyuanZhao: I don't think we could cache the flux, since that may be huge. Last time we checked with Mark he said we could just throw an error since result sets aren't meant to be mapped through once than once.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
ChengyuanZhao: oh yea!! perfect!

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
dzou: Cool, got it.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
dzou: Edit: it might actually be a `.map` instead of flatmap.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
ChengyuanZhao: yes!

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
ChengyuanZhao: done

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -95,6 +94,9 @@ public Statement bindNull(int i, Class<?> type) {
     return Mono
         .just(new SpannerResult(
             Flux.create(sink -> result.subscribe(new PartialResultFluxConverter(sink))),
-            Mono.create(sink -> result.subscribe(new PartialResultSetStatsConverter(sink)))));
+            result.next().flatMap(partialResultSet -> partialResultSet.hasStats() ? Mono
```
dzou: Since getting the stats out of the partialResultSet is not an async operation, you can just use a simple map:

```
result.next().map(partialResultSet -> {
  if (partialResultSet.hasStats()) {
     return Math.tointExact...blahblah
  } else {
     return 0;
  });
```

ChengyuanZhao: yes fixed!

src/test/java/com/google/cloud/spanner/r2dbc/SpannerStatementTest.java
```diff
@@ -69,4 +83,52 @@ public void executeDummyImplementation() {
...
+        .addValues(this.a1).build();
+
+    PartialResultSet p2 = PartialResultSet.newBuilder().setChunkedValue(false)
+        .addValues(this.a2).build();
+
+    Flux<PartialResultSet> inputs = Flux.just(p1, p2);
+
+    when(this.mockClient.executeStreamingSql(any(), any(), any())).thenReturn(inputs);
+
+    assertThat(Mono.from(new SpannerStatement(this.mockClient, null, null, null).execute())
```
elefeint: Can you also assert that the rows returned are as expected? I have paranoia that .next() will drain a value from rows flux when returning the mono of stats.

ChengyuanZhao: done.

PR #68, opened 2019-05-22
Upgrade to SPI 0.8, add missing interface methods

### LINKED ISSUE #58: Upgrade to R2DBC 0.8.0 M8
Changelog: https://r2dbc.io/spec/0.8.0.M8/CHANGELOG.txt
Javadoc: https://r2dbc.io/spec/0.8.0.M8/api/
Spec: https://r2dbc.io/spec/0.8.0.M8/spec/html/

elefeint:
The upgrade added a couple of trivial interface methods.

While I was there, I folded everything under a single for-loop in `SpannerRowMetadata`. It already had a for-loop and two stream pipeline. Since now there are three different pieces of data we need to keep track of, a single for-loop is simpler.

Fixes #58 .

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
-    <r2dbc.version>1.0.0.M7</r2dbc.version>
+    <r2dbc.version>0.8.0.M8</r2dbc.version>
     <reactor.version>Californium-SR6</reactor.version>
     <google-auth.version>0.15.0</google-auth.version>
     <grpc.version>1.20.0</grpc.version>
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerRowMetadata.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerRowMetadata.java
@@ -38,14 +38,17 @@ public class SpannerRowMetadata implements RowMetadata {
   /**
    * Mapping of column names to its integer index position in the row.
    */
-  private final HashMap<String, Integer> columnNameIndex = new HashMap<>();
+  private final HashMap<String, Integer> columnNameIndex;
 
   /**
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -14,7 +14,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
-    <r2dbc.version>1.0.0.M7</r2dbc.version>
+    <r2dbc.version>0.8.0.M8</r2dbc.version>
```
meltsufin: On the surface, looks like a downgrade.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerRowMetadata.java
```diff
@@ -32,27 +33,32 @@
 
   private final List<ColumnMetadata> columnMetadatas;
 
+  private final List<String> columnNames;
+
   /**
    * Mapping of column names to its integer index position in the row.
    */
-  private final HashMap<String, Integer> columnNameIndex;
+  private final HashMap<String, Integer> columnNameIndex = new HashMap<>();
```
meltsufin: What was wrong with initialization in the constructor?

pom.xml
elefeint: The naming convention changed. I recall a conversation that R2DBC was going to release as 0.8 until the feature set solidified. So, there would be an 0.8 GA.

Here is the change on SPI side: https://github.com/r2dbc/r2dbc-spi/commit/e178e59ceaf27f55649ce684c97f1145fe8f3f06

src/main/java/com/google/cloud/spanner/r2dbc/SpannerRowMetadata.java
elefeint: Nothing; I'll put it back. I default to field initialization when there is no specific reason for initializing in constructor, but there is no particular reason for it. I can see the argument for being consistent, and initializing everything in the constructor, too.

PR #71, opened 2019-05-22
Simplify Partial ResultSet Assembly

dzou:
This PR simplifies the PartialResultSet assembly by replacing ConvertingFluxAdapter with `Flux.flatMapIterable`.

[Flux.flatMapIterable](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#flatMapIterable-java.util.function.Function-) actually does the equivalent thing to the ConvertingFluxAdapter (i.e. mapping one element to zero-to-N elements and emitting elements as a flux).

Also changes logic to use switchOnFirst; this will make it so that only one _hot_ flux is available in a SpannerResult at any time.


src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -87,15 +86,26 @@ public Statement bindNull(int i, Class<?> type) {
...
+    PartialResultRowExtractor partialResultRowExtractor = new PartialResultRowExtractor();
+
+    return this.client.executeStreamingSql(this.session, this.transaction, this.sql)
+        .switchOnFirst((signal, flux) -> {
+          if (signal.hasError()) {
+            return flux.then(Mono.empty());
+          }
+
+          PartialResultSet firstPartialResultSet = signal.get();
+          if (firstPartialResultSet.hasStats()) {
```
ChengyuanZhao: What if the first partial result set is also the only/last partial result set? Wouldn't that mean the stats ( which encompasses more than just rows changed for a read query) is also part of the first partial result set? 
Wouldn't that mean we then give an empty flux of rows?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -87,15 +86,26 @@ public Statement bindNull(int i, Class<?> type) {
...
+                flux.next().map(partialResultSet ->
+                    Math.toIntExact(partialResultSet.getStats().getRowCountExact()));
+            return Mono.just(new SpannerResult(Flux.empty(), rowsChanged));
+          } else {
+            return Mono.just(new SpannerResult(
+                flux.flatMapIterable(partialResultRowExtractor),
+                Mono.just(0)));
+          }
+        })
+        .next();
```
ChengyuanZhao: Hmm I'm not sure the readability has been improved or if the actual logic was simplified. Indeed you do delete a class, which is nice ,but it was a very simple and readable class. 

I think what's here is somewhat less simple actually. Especially because the final returned mono of SpanneResult comes from a .next() call on line 109, which (while totally logical) is just a bit harder to read.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
dzou: Got it. Updated the rowCount checking condition to check rowCountExact and rowCountLowerBound. This should be the correct check to see if a PRS corresponds to update vs. a read.

https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#resultsetstats

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
dzou: As discussed offline, I think this approach is better because now only a single flux/publisher is available in a result, which resolves the issue of a query being run twice if the user calls `.map()` and `.getRowsUpdated()`.

dzou:
> I like that new test you added. It probably is the intended behavior now!

Cool. PTAL, I added that additional check to the row emitter we discussed in the latest commit.
```
 if (partialResultSet.getValuesList().isEmpty()) {
      return Collections.emptyList();
}
```

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -87,15 +86,35 @@ public Statement bindNull(int i, Class<?> type) {
...
+          if (signal.hasError()) {
+            return flux.then(Mono.empty());
+          }
+
+          PartialResultSet firstPartialResultSet = signal.get();
+          if (hasUpdatedRowCount(firstPartialResultSet)) {
+            Mono<Integer> rowsChanged =
+                flux.next().map(partialResultSet ->
+                    Math.toIntExact(partialResultSet.getStats().getRowCountExact()));
+            return Mono.just(new SpannerResult(Flux.empty(), rowsChanged));
```
elefeint: So this enables us to eventually use a different class to represent DML results... Because row count is not relevant to select queries, and row stream is not relevant to update queries. Very cool!

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -87,15 +86,35 @@ public Statement bindNull(int i, Class<?> type) {
...
+                flux.flatMapIterable(partialResultRowExtractor),
+                Mono.just(0)));
+          }
+        })
+        .next();
+  }
+
+  private static boolean hasUpdatedRowCount(PartialResultSet firstPartialResultSet) {
+    if (firstPartialResultSet.hasStats()) {
+      return firstPartialResultSet.getStats().getRowCountExact() > 0
```
elefeint: Hmm, what if it's an update query that happened to affect 0 rows? Could you test what happens on, say, "update row where key='does not exist'"? With the current logic, it would be treated like a select query, not sure if it would end up outputting a fake row of 0 columns.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
dzou: Yup, the switchOnFirst lets us conditionally modify stream based on first emitted partial result set; I think this is a promising approach for letting us choose the correct constructor based on the first PRS's contents.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
dzou: Nice catch; updated this with `RowCountCase` which should now accurately distinguish btwn update with 0 rows and an unset field. And added a integration test.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -87,15 +87,31 @@ public Statement bindNull(int i, Class<?> type) {
...
-                .subscribe(new ConvertingFluxAdapter(sink, new PartialResultRowExtractor()))),
-            result.next().map(partialResultSet -> partialResultSet.hasStats()
-                ? Math.toIntExact(partialResultSet.getStats().getRowCountExact())
-                : 0)));
+    PartialResultRowExtractor partialResultRowExtractor = new PartialResultRowExtractor();
+
+    return this.client.executeStreamingSql(this.session, this.transaction, this.sql)
+        .switchOnFirst((signal, flux) -> {
+          if (signal.hasError()) {
+            return flux.then(Mono.empty());
```
elefeint: We should propagate the error:
`Mono.error(signal.getThrowable())`

src/test/java/com/google/cloud/spanner/r2dbc/SpannerStatementTest.java
```diff
@@ -139,4 +141,31 @@ public void readDmlQueryTest() {
         .flatMap(r -> Mono.from(r.getRowsUpdated())).block()).isEqualTo(555);
   }
 
+  @Test
+  public void noopMapOnUpdateQueries() {
+    Client mockClient = mock(Client.class);
+    String sql = "delete from Books where true";
+    PartialResultSet partialResultSet = PartialResultSet.newBuilder()
+        .setMetadata(ResultSetMetadata.newBuilder().setRowType(StructType.newBuilder()
```
elefeint: Would a result set from a delete operation even contain the column description? Creation of `StructType` might not be needed here.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerStatementTest.java
```diff
@@ -139,4 +141,31 @@ public void readDmlQueryTest() {
         .flatMap(r -> Mono.from(r.getRowsUpdated())).block()).isEqualTo(555);
   }
 
+  @Test
+  public void noopMapOnUpdateQueries() {
```
elefeint: Could you add a suffix `WhenNoRowsAffected` to the test, so it's easy to see why it's a no-op?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
dzou: Done.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerStatementTest.java
dzou: Done.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerStatementTest.java
dzou: Done.

Issue #55, opened 2019-05-17
Replace blocking with StepVerifier in tests


elefeint:
Step verifier is more flexible, and will lead to more readable tests as the tests get more complex.

https://projectreactor.io/docs/core/release/reference/#_testing_a_scenario_with_stepverifier
https://projectreactor.io/docs/test/release/api/reactor/test/StepVerifier.html

JoeWang1127:
close as not planned

PR #73, opened 2019-05-23
Enhance test to verify that the innermost mono returned by client actually runs

### LINKED ISSUE #49: Update SpannerConnectionTest to fail if implementation flatMap() are replaced by map()
https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/41#pullrequestreview-237291024

Or prove it should not fail.

map() leads to double wrapped Mono. The current test blocks on the outer Mono, so it might verify despite results not yet being available.

elefeint:
Fixes #49.

### DIFF
--- a/src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
+++ b/src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
@@ -35,6 +35,8 @@ import com.google.spanner.v1.Transaction;
 import com.google.spanner.v1.Type;
 import com.google.spanner.v1.TypeCode;
 import io.r2dbc.spi.Statement;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mockito;
@@ -92,6 +94,15 @@ public class SpannerConnectionTest {
   public void beginAndCommitTransactions() {
     SpannerConnection connection = new SpannerConnection(this.mockClient, TEST_SESSION);
 
+    AtomicInteger beginTransactionCounter = new AtomicInteger(0);
+    AtomicInteger commitTransactionCounter = new AtomicInteger(0);
+    when(this.mockClient.beginTransaction(TEST_SESSION))
+        .thenReturn(this.getTrackingMono(
+            () -> Transaction.getDefaultInstance(), beginTransactionCounter));
+    when(this.mockClient.commitTransaction(TEST_SESSION,  Transaction.getDefaultInstance()))
+        .thenReturn(this.getTrackingMono(
+            () -> CommitResponse.getDefaultInstance(), commitTransactionCounter));
+
     Mono.from(connection.commitTransaction()).block();
     verify(this.mockClient, never()).commitTransaction(any(), any());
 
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
```diff
@@ -116,5 +140,22 @@ public void rollbackTransactions() {
...
+    return Mono.fromSupplier(() -> {
+      counter.addAndGet(1);
+      return returnObjectSupplier.get();
+    });
+  }
+
+  private void verifyTrackingMono(String description, AtomicInteger counter) {
+    assertThat(counter.get())
+        .withFailMessage(description + " publisher was never subscribed to.")
+        .isEqualTo(1);
```
dzou: I think this method could be more clear if the expected count was passed in via parameter; I would also rename the method to maybe `verifyCount` or something similar since it does not accept a mono as a parameter but atomic int instead.

Alternatively, I would also be open to just unrolling the helper function and adding the two assert statements inline since it is only repeated twice.

elefeint: yeah, I think I'll do the latter -- fold the assertions into each test.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
```diff
@@ -92,6 +94,15 @@ public void executeStatementReturnsWorkingStatementWithCorrectQuery() {
...
+    AtomicInteger beginTransactionCounter = new AtomicInteger(0);
+    AtomicInteger commitTransactionCounter = new AtomicInteger(0);
+    when(this.mockClient.beginTransaction(TEST_SESSION))
+        .thenReturn(this.getTrackingMono(
+            () -> Transaction.getDefaultInstance(), beginTransactionCounter));
+    when(this.mockClient.commitTransaction(TEST_SESSION,  Transaction.getDefaultInstance()))
+        .thenReturn(this.getTrackingMono(
+            () -> CommitResponse.getDefaultInstance(), commitTransactionCounter));
+
     Mono.from(connection.commitTransaction()).block();
```
meltsufin: Please add a comment saying what this block is testing... "no-op commit when connection is not started".

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
```diff
@@ -101,12 +112,29 @@ public void beginAndCommitTransactions() {
         .beginTransaction(TEST_SESSION);
     verify(this.mockClient, times(1))
         .commitTransaction(TEST_SESSION, Transaction.getDefaultInstance());
+
+    assertThat(beginTransactionCounter.get())
+        .withFailMessage("beginTransaction publisher was never subscribed to.")
+        .isEqualTo(1);
```
meltsufin: I think it reads more fluently if you put `isEqualTo` before `withFailMessage`.

elefeint:
Yes, I have -- there is a tracking issue for it, #55.

PublisherProbe can be useful, although it won't reduce boilerplate that much.

meltsufin:
@elefeint Do you want to try using the `reactor-testing` methods as part of this PR, or leave it for later?

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
elefeint: I'll just split it off into a small test for only that scenario, since a noop commit does not require any of the other verification in this test.

PR #78, opened 2019-05-24
Improve detection of DML vs Read queries

### LINKED ISSUE #75: Handle parsing Stats for Partitioned DML Queries
For Partitioned DML queries, the stats of the query is found in the second partial result set returned in the stream.

You can see below, this is a list of the two PRS returned from executing a Partitioned DML query:

image

The PRS emitted in the stream is the one that contains the stats.

This means we cannot rely on our existing strategy of assuming that the first PRS always has the stats for partitioned DML queries.

### LINKED ISSUE #77: Query never finishes if there are no matching rows
To replicate the issue, just put this code into `SpannerIT.testQuerying()`
```
    List<String> result = Mono.from(this.connectionFactory.create())
        .map(connection -> connection.createStatement("SELECT title, author FROM books where author = 'Stephen King'"))
        .flatMapMany(statement -> statement.execute())
        .flatMap(spannerResult -> spannerResult.map(
            (r, meta) -> r.get(0, String.class) + " by " + r.get(1, String.class)
        ))
        .doOnNext(s -> System.out.println("Book: " + s))
        .collectList()
        .block();
```

dzou:
This improves the detection of DML vs. Read queries by checking if the metadata's field list is empty or not; this appears to be the only reliable indicator to distinguish between the two types.

Fixes #77.
Fixes #75.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
@@ -19,7 +19,6 @@ package com.google.cloud.spanner.r2dbc;
 import com.google.cloud.spanner.r2dbc.client.Client;
 import com.google.cloud.spanner.r2dbc.result.PartialResultRowExtractor;
 import com.google.spanner.v1.PartialResultSet;
-import com.google.spanner.v1.ResultSetStats.RowCountCase;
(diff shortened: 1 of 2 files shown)


dzou:
> Maybe we should ask the Spanner team to confirm that this is a reliable way to detect DML query result.

Will do.

PR #80, opened 2019-05-28
Set Transaction as Volatile Variable in Connection

dzou:
As per discussion today with R2DBC folks, this PR modifies `SpannerConnection` to set the transaction as a volatile property of the connection class.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
@@ -39,7 +39,7 @@ public class SpannerConnection implements Connection {
 
   private final Session session;
 
-  private Mono<Transaction> currentTransaction;
+  private volatile Transaction currentTransaction;
 
   /**
    * Instantiates a Spanner session with given configuration.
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -49,33 +49,38 @@
...
   @Override
   public Publisher<Void> rollbackTransaction() {
-    return this.currentTransaction
-        .flatMap(transaction -> this.client.rollbackTransaction(this.session, transaction))
-        .switchIfEmpty(Mono.fromRunnable(() ->
-          this.logger.warn("rollbackTransaction() is a no-op; called with no transaction active.")))
-        .then();
+    return Mono.defer(() -> {
+      if (this.currentTransaction == null) {
+        this.logger.warn("commitTransaction() is a no-op; called with no transaction active.");
```
elefeint: "rollbackTransaction"

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -162,28 +163,34 @@ public GrpcClient(GoogleCredentials credentials) {
...
-        .map(t ->  ExecuteSqlRequest.newBuilder()
+      Session session, @Nullable Transaction transaction, String sql) {
+
+    return Flux.create(sink -> {
+      TransactionSelector transactionSelector;
+      if (transaction != null) {
+        transactionSelector =
+            TransactionSelector.newBuilder().setId(transaction.getId()).build();
+      } else {
+        transactionSelector = readOnlySingleUseTransaction();
```
elefeint: Since we can now use imperative style here (if transaction is an already-available object, and not a deferred mono), `readOnlySingleUseTransaction()` is not needed at all. The default behavior when you omit a transaction from `ExecuteSqlRequest` is to assume a single-use read-only strong-consistency transaction.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -39,7 +39,7 @@
 
   private final Session session;
 
-  private Mono<Transaction> currentTransaction;
```
dzou: @mp911de - Hey Mark, we're following your advice from the meeting today about setting the Transaction object as `volatile`. We wanted to circle back with you to ask you about this approach again - is there a functionality advantage to this approach or was it more about the style?

We were wondering if the previous approach was superior from functionality point-of-view. You can see here we stored the `Mono<Transaction>` as the variable here; this allowed us to chain the `commitTransaction()` to the result of `Mono<Transaction>`. In the volatile approach we are no longer able to chain `commit` and `rollback` to the `Mono<Transaction>`.

But yeah, unsure if it makes a difference.

mp911de: There's a slight performance advantage if you keep the item in a `volatile` field. For now, this is entirely a style issue.

With r2dbc/r2dbc-spi#95 ([`Connection.isAutoCommit()`](https://github.com/r2dbc/r2dbc-spi/pull/95/files#diff-bb279211f4e3cbfdafd443ffbcb3706aR92)) you will need a scalar representation of the transaction state anyway, that's why I mentioned the issue at all.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
dzou: Donerino.

PR #81, opened 2019-05-28
Modify Retry Approach to Propagate Errors

dzou:
This modifies the retry logic of the driver to propagate `R2dbcTransientExceptions` if a call fails instead of attempting to retry on the driver-side.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
@@ -40,11 +40,7 @@ import io.grpc.CallCredentials;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.auth.MoreCallCredentials;
-import io.grpc.stub.ClientCallStreamObserver;
-import io.grpc.stub.ClientResponseObserver;
-import java.io.IOException;
 import reactor.core.publisher.Flux;
-import reactor.core.publisher.FluxSink;
 import reactor.core.publisher.Mono;
 
 /**
@@ -83,7 +79,7 @@ public class GrpcClient implements Client {
   }
 
   @VisibleForTesting
-  GrpcClient(SpannerStub spanner) throws IOException {
+  GrpcClient(SpannerStub spanner) {
     this.spanner = spanner;
     this.channel = null;
   }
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
```diff
@@ -88,6 +79,10 @@ public void onNext(ResponseT value) {
 
     @Override
     public void onError(Throwable t) {
+      if (isRetryable(t)) {
+        t = new R2dbcTransientResourceException(t);
+      }
```
meltsufin: else wrap it into `R2dbcNonTransientException`?

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
```diff
@@ -135,6 +130,11 @@ public void onNext(ResponseT response) {
     @Override
     public void onError(Throwable throwable) {
       this.terminalEventReceived = true;
+
+      if (isRetryable(throwable)) {
+        throwable = new R2dbcTransientResourceException(throwable);
+      }
```
meltsufin: Same here wrap into `R2dbcNonTransientException`?
Also, consider adding an exception reason.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -171,60 +167,9 @@ public GrpcClient(GoogleCredentials credentials) {
...
-                  .request((int) Math.min(demand, Integer.MAX_VALUE)));
-          }));
-  }
-
-  @VisibleForTesting
-  SpannerStub getSpanner() {
-    return this.spanner;
-  }
-
-  private static final class SinkResponseObserver<ReqT, RespT> implements
```
meltsufin: Nice catch! How did we miss this refactor? Probably tricky merges.

elefeint: I originally [meant to remove](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/blob/246e6f656ae748a8061af13f3e62b34378dd4009/src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java#L59) the implementation in ObservableReactiveUtil, but forgot.
They are functionally equivalent, I think, so it does not matter which one to keep.

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
dzou: Done.

src/main/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtil.java
dzou: Done.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java

dzou:
> If you are keeping the ObservableReactiveUtil implementation, can you remove my comment about "this will go away"?
> Otherwise LGTM; I agree with Mike's suggestion to wrap in R2dbcNonTransientException for the other case.

Gotcha. Donerino.

src/test/java/com/google/cloud/spanner/r2dbc/util/ObservableReactiveUtilTest.java
```diff
@@ -45,57 +47,51 @@ public void unaryCallForwardsError() {
       observer.onError(new IllegalArgumentException("oh no"));
     });
     assertThatThrownBy(() -> mono.block())
-        .isInstanceOf(IllegalArgumentException.class)
```
elefeint: Could you keep the check for root exception as `hasCauseInstanceOf()`

dzou: Done.

PR #87, opened 2019-05-30
Statement parameter binding

### LINKED ISSUE #13: Implement parameter binding
in `SpannerStatement`.

dmitry-s:
fixes #13


src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerIT.java
```diff
@@ -122,15 +123,29 @@ public void testQuerying() {
...
-    executeDmlQuery(
-        "INSERT BOOKS (UUID, TITLE, AUTHOR, CATEGORY) VALUES"
-            + " ('df0e3d06-2743-4691-8e51-6d33d90c5cb9', 'Effective Java', "
-            + "'Joshua Bloch', 100)");
-
-    executeDmlQuery(
-        "INSERT BOOKS (UUID, TITLE, AUTHOR, CATEGORY) VALUES"
-            + " ('2b2cbd78-ecd8-430e-b685-fa7910f8a4c7', 'JavaScript: "
-            + "The Good Parts', 'Douglas Crockford', 100);");
+    Mono.from(this.connectionFactory.create())
```
ChengyuanZhao: @meltsufin @elefeint @dzou Should we make an issue for them asking why they insist on making all the return types in the SPI "publisher" instead of actually Mono or Flux? 

Even in their own examples they're endlessly wrapping them in .from() calls.....

elefeint: No, this is an understandable design choice. Reactive Streams is a public specification, built into Java9+. Project Reactor is just one framework implementing the spec, so tying R2DBC to it would limit options for driver implementors and unnecessarily tie their success paths.

In Spanner R2DBC driver, we've made a choice to depend on Project Reactor, so we can definitely return Flux/Mono from all our implementations of R2DBC SPI (they are covariant, so no problem).

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -55,45 +75,88 @@
...
+    return this;
   }
 
   @Override
-  public Statement bind(Object o, Object o1) {
-    return null;
+  public Statement bind(Object identifier, Object value) {
+    requireNonNull(identifier);
+    if (identifier instanceof String) {
+      this.bindings.getLast().put((String)identifier, value);
```
elefeint: LinkedList.getLast() is efficient, but if we keep a map of in-progress/incomplete bindings separately, then there is no need for getting it, and also no need to add an empty list in constructor.
Then `add()` would add the incomplete map (that has just become complete) and get a new map going.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -55,45 +75,88 @@
...
+      if (types != null) {
+        this.types = types;
+      }
+
+      paramsStructs.add(paramsStructBuilder.build());
+    }
+
+    Flux<Struct> structFlux = Flux.fromIterable(paramsStructs);
+
+    if (this.sql != null && this.sql.trim().toLowerCase().startsWith("select")) {
```
elefeint: Could you add assertion for `sql` being non-null to the constructor? A statement with null sql would not be valid, so we should fail early on that.

elefeint: I know it's trivial logic, but can the second half of the condition go into a `isSelectQuery` helper method?

src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerIT.java
ChengyuanZhao: I see I see.  Seems like my issue is more with the lack of functionality in the Publisher interface then :/

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -55,45 +75,88 @@
...
+      return structFlux.flatMap(this::runSingleStatement);
+    }
+    return structFlux.concatMapDelayError(this::runSingleStatement);
```
elefeint: Could you add a comment about Spanner updates required to be in order?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -55,45 +75,88 @@
...
   }
 
   @Override
   public Publisher<? extends Result> execute() {
+    List<Struct> paramsStructs = new ArrayList<>();
+    for (Map<String, Object> bindingsBatch : this.bindings) {
+      Builder paramsStructBuilder = Struct.newBuilder();
+      Map<String, Type> types = this.types.isEmpty() ? new HashMap<>() : null;
+
+      for (Map.Entry<String, Object> binding : bindingsBatch.entrySet()) {
```
elefeint: Would it make sense to put the logic in a helper method and call it from `.add()`? We are building up a map, might as well be building up a list of `Struct`s.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -55,45 +75,88 @@
...
+        String paramName = binding.getKey();
+        Codec codec = this.resolvedCodecs.computeIfAbsent(paramName,
```
elefeint: The same binding for all rows should be of the same type, right? Can we save time by only doing this on the first `add()`?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -37,16 +50,23 @@
 
   private Session session;
 
-  private Transaction transaction;
+  private SpannerTransactionContext transaction;
 
   private String sql;
 
+  private LinkedList<Map<String, Object>> bindings = new LinkedList<>();
```
dzou: I think in practice ArrayList is typically preferred over LinkedList because of caching advantages.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -55,45 +75,88 @@
...
   public Statement bindNull(int i, Class<?> type) {
-    return null;
+    throw new IllegalArgumentException("Only named parameters are supported");
   }
 
   @Override
   public Publisher<? extends Result> execute() {
+    List<Struct> paramsStructs = new ArrayList<>();
+    for (Map<String, Object> bindingsBatch : this.bindings) {
+      Builder paramsStructBuilder = Struct.newBuilder();
```
dzou: minor style thing - it would be better to qualify this `Builder` class with `Struct.Builder`

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
dzou: Yeah I would recommend this approach too. You could keep a `List<Struct.Builder>` instead of Map which will save a layer of conversions; then just build them all when you are ready to execute the query.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -155,20 +159,27 @@ public GrpcClient(GoogleCredentials credentials) {
...
               .setSql(sql)
               .setSession(session.getName());
+      if (params != null) {
+        executeSqlRequest
+            .setParams(params)
+            .putAllParamTypes(types);
+      }
 
-      if (transaction != null) {
+      if (transactionContext != null && transactionContext.getTransaction() != null) {
```
dzou: I think you should just add a `Assert.notNull(transaction)` in the constructor of TransactionContext, and then simplify this to just check `transactionContext != null`. This way users in the future will not have to worry about doing 2 null-checks for transactions.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerStatementTest.java
```diff
@@ -73,19 +80,76 @@ public void executeDummyImplementation() {
...
-    result.block().map((r, m) -> (String)r.get(0)).blockFirst().equals("Odyssey");
+    assertThat(result.next().block().map((r, m) -> (String)r.get(0)).blockFirst())
+        .isEqualTo("Odyssey");
 
-    verify(mockClient).executeStreamingSql(TEST_SESSION, null, sql);
+    verify(mockClient).executeStreamingSql(eq(TEST_SESSION), isNull(), eq(sql), any(), any());
+  }
+
+  @Test
+  public void executeDummyImplementationBind() {
```
meltsufin: Can you add some comments in this method to separate blocks that are setting up the dummy values and mocks vs. where the actual method under test is executed? These reactive tests are a pain to read. Hopefully, some comments would help.

src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerIT.java
```diff
@@ -94,6 +101,34 @@ public void setupStubs() throws IOException {
...
         .withCallCredentials(callCredentials);
+
+    SpannerOptions options = SpannerOptions.newBuilder().build();
+    Spanner spanner = options.getService();
+
+    DatabaseId id = DatabaseId.of(options.getProjectId(), TEST_INSTANCE, TEST_DATABASE);
+
+    DatabaseAdminClient dbAdminClient = spanner.getDatabaseAdminClient();
+
+    dbAdminClient.updateDatabaseDdl(
```
meltsufin: This is not exactly setting up the stubs. :)
Rename the method and/or put this in a separate method?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -55,45 +75,88 @@
...
+      this.bindings.getLast().put((String)identifier, value);
+      return this;
+    }
+    throw new IllegalArgumentException("Only String identifiers are supported");
   }
 
   @Override
   public Statement bind(int i, Object o) {
-    return null;
+    throw new IllegalArgumentException("Only named parameters are supported");
```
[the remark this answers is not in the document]
dmitry-s: I thought we decided that we don't want to support that, because it would require parsing of the SQL string.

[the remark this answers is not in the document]
meltsufin: Either way, let's track this as an issue.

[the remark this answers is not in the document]
ChengyuanZhao: https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/issues/91

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
ChengyuanZhao: These aren't actually columns if i'm not mistaken. These are param tags in the query string, which means they don't need to be the same type. For example , you can get the size in bytes in googleSQL of a STRING or BYTES, so that single tag could take on either value. 

Furthermore, I think the codec can change too? For example, if the tag in the SQL is like `where 100 < @tag` then could two different codecs convert either a long or a string into a number to fill that tag?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
ChengyuanZhao: Yea I think that might complicate things. Especially since the user doesn't need to call add() to close/finalize the final one. That would mean some sort of code in the execute() that would add on this final map, which feels out of place to be in in the execute() method.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -113,45 +125,28 @@ public Statement bindNull(Object identifier, Class<?> type) {
...
-    if (this.sql != null && this.sql.trim().toLowerCase().startsWith("select")) {
+    if (isSelectQuery()) {
       return structFlux.flatMap(this::runSingleStatement);
     }
+    //DML statements have to be executed sequentially because they need seqNo to be in order
     return structFlux.concatMapDelayError(this::runSingleStatement);
   }
 
+  private boolean isSelectQuery() {
+    return this.sql.trim().toLowerCase().startsWith("select");
```
elefeint: optimization for the future: only lowercasing the prefix. Does not need to be in this PR.

elefeint:
Integration test is unhappy tho:
```
SEVERE: *~*~*~ Channel ManagedChannelImpl{logId=41, target=spanner.googleapis.com:443} was not shutdown properly!!! ~*~*~*
    Make sure to call shutdown()/shutdownNow() and wait until awaitTermination() returns true.
java.lang.RuntimeException: ManagedChannel allocation site
```

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -55,45 +74,84 @@
...
+        this.currentBindingsBuilder = Struct.newBuilder();
+      }
+      this.currentBindingsBuilder.putFields(paramName, codec.encode(value));
+      if (this.bindingsStucts.isEmpty()) {
+        //first binding, fill types map
+        this.types.put(paramName, Type.newBuilder().setCode(codec.getTypeCode()).build());
+      }
+      return this;
+    }
+    throw new IllegalArgumentException("Only String identifiers are supported");
```
dzou: You could make this into a helper in the `Assert` helper; i.e. `Assert.requireInstanceOf(..)`

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -55,45 +74,84 @@
...
   }
 
   @Override
-  public Statement bind(Object o, Object o1) {
-    return null;
+  public Statement bind(Object identifier, Object value) {
+    requireNonNull(identifier);
+    if (identifier instanceof String) {
+      String paramName = (String) identifier;
+      //we assume all parameters with the same name have the same type
```
meltsufin: style nit: space before start of comment text
```suggestion
      // we assume all parameters with the same name have the same type
```

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -55,45 +74,84 @@
...
+      Codec codec = this.resolvedCodecs
+          .computeIfAbsent(paramName, n -> this.codecs.getCodec(value));
+      if (this.currentBindingsBuilder == null) {
+        this.currentBindingsBuilder = Struct.newBuilder();
+      }
+      this.currentBindingsBuilder.putFields(paramName, codec.encode(value));
+      if (this.bindingsStucts.isEmpty()) {
+        //first binding, fill types map
```
meltsufin:
```suggestion
        // first binding, fill types map
```

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -55,45 +74,84 @@
...
+    add();
+    if (this.bindingsStucts.size() == 0) {
+      this.bindingsStucts.add(Struct.newBuilder().build());
+    }
+    Flux<Struct> structFlux = Flux.fromIterable(this.bindingsStucts);
+
+    if (isSelectQuery()) {
+      return structFlux.flatMap(this::runSingleStatement);
+    }
+    //DML statements have to be executed sequentially because they need seqNo to be in order
```
meltsufin:
```suggestion
    // DML statements have to be executed sequentially because they need seqNo to be in order
```

PR #95, opened 2019-06-04
Control result rows demand, requesting one PartialResultSet by default

### LINKED ISSUE #9: Investigate changes needed in Client Library to support backpressure
Right now, client library requests one more partial result set whenever it processes the previous result set. 
This may well be the correct behavior, given that Spanner PartialResultSet does not correspond to a row. But perhaps this can be made more efficient by providing hints to GrpcResultSet for the number of rows to retrieve next.

elefeint:
`FlatMapIterable` keeps a buffer of a specified size. Because we cannot predict the relationship between a `PartialResultSet` and a spanner row, we'll default to requesting one `PartialResultSet` at a time.

Clients can modify the default batch size by setting a property, either on connection factory level (`partial_result_set_prefetch`) or on a specific Statement (`setPartialResultSetPrefetch`).

Fixes #9. Sort of.


src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -105,11 +118,15 @@ public SpannerConnectionConfiguration build() throws IOException {
...
       }
 
-      return new SpannerConnectionConfiguration(
+      SpannerConnectionConfiguration configuration = new SpannerConnectionConfiguration(
           this.projectId,
           this.instanceName,
           this.databaseName,
           this.credentials);
+
+      configuration.partialResultSetPrefetch = this.partialResultSetPrefetch;
```
dzou: Should this param just be added to the constructor of SpannerConnectionConfiguration for consistency? Instead of directly setting the field; that way you can also keep `partialResultSetPrefetch` final.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
```diff
@@ -92,4 +87,23 @@ void setClient(Client client) {
     this.client = client;
   }
 
+  private SpannerConnectionConfiguration createConfiguration(ConnectionFactoryOptions options) {
```
dzou: This method can be made static.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -32,6 +32,8 @@
 
   private final GoogleCredentials credentials;
 
+  private int partialResultSetPrefetch = 1;
```
meltsufin: `partialResultSetFetchSize`?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -77,6 +83,8 @@ public GoogleCredentials getCredentials() {
 
     private GoogleCredentials credentials;
 
+    private int partialResultSetPrefetch = 1;
```
meltsufin: I think it should be `Integer`. Otherwise, you're duplicating the default value in the builder and the actual configuration class where the field is optional.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
meltsufin: conditional on whether it was set

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
```diff
@@ -46,6 +46,9 @@
   /** Option name for GCP Spanner instance. */
   public static final Option<String> INSTANCE = Option.valueOf("instance");
 
+  public static final Option<Integer> PARTIAL_RESULT_SET_PREFETCH
+      = Option.valueOf("partial_result_set_prefetch");
```
meltsufin: maybe `partial_result_set_fetch_size`?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
```diff
@@ -92,4 +87,23 @@ void setClient(Client client) {
...
+          .setDatabaseName(options.getRequiredValue(DATABASE))
+          .setCredentials(options.getValue(GOOGLE_CREDENTIALS));
+
+      if (options.hasOption(PARTIAL_RESULT_SET_PREFETCH)) {
+        configBuilder.setPartialResultSetPrefetch(
+            options.getValue(PARTIAL_RESULT_SET_PREFETCH));
+      }
+      return configBuilder.build();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
meltsufin: What's causing the `IOException`? Can we please add an exception message here?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -41,6 +41,8 @@
 
   private String sql;
 
+  private int partialResultSetPrefetch = 1;
```
meltsufin: Duplicating  this default yet again? In fact, this is probably the only place you need the default of `1` to exist.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -98,6 +100,9 @@ public Statement bindNull(int i, Class<?> type) {
           if (signal.hasError()) {
             return Mono.error(signal.getThrowable());
           }
+          if (signal.isOnComplete()) {
+            return Mono.just(new SpannerResult(Flux.empty(), Mono.just(0)));
```
dzou: For this line, I saw in the javadoc for switchOnFirst that they have a caveat:
> It is advised to return a Publisher derived from the original Flux in all cases, as not doing so would keep the original Publisher open and hanging with a single request until the inner Publisher terminates or the whole Flux is cancelled.

https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#switchOnFirst-java.util.function.BiFunction-

Will returning a mono not based on the original flux be problematic? I think it should be okay if its already complete...? Otherwise, you could something like `flux.then(Mono.just(new SpannerResult(..)));`

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
elefeint: I was thinking it was messy to keep setting it, but then the alternative is creating a wrapper Integer object just to have the privilege of leaving it null...

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -100,7 +103,10 @@ public Batch createBatch() {
 
   @Override
   public Statement createStatement(String sql) {
-    return new SpannerStatement(this.client, this.session, this.currentTransaction, sql);
+    SpannerStatement statement
```
dzou: Should SpannerStatement also be made to simply accept the SpannerConnectionConfiguration object through constructor, rather than accept the prefetch through a setter? similar to how connection has access to the configuration.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
elefeint: Yeah. It's acquiring default credentials, since that involves the file system. Kind of annoying to have it here in the first place; that's why we discussed a provider at some point.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProviderTest.java
```diff
@@ -95,6 +110,61 @@ public void getDriverReturnsSpanner() {
     assertThat(this.spannerConnectionFactoryProvider.getDriver()).isEqualTo(DRIVER_NAME);
   }
 
+  @Test
+  public void partialResultSetPrefetchPropagatesAsDemand() {
```
dzou: Nice test :+1:

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
elefeint: That ambiguous "set"... Damn you, English, for being so concise!

elefeint: PartialResult SetFetchSize
PartialResultSet FetchSize
prsFetchSize is too cryptic...
Halp.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
elefeint: I'm going to move it closer to where the problem first occurs, and wrap it into a non-transient reosource exception.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
meltsufin: setPartialResultSetSize and getPartialResultSetSize seem fine to me.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
elefeint: Setters are convenient for optional properties; then the constructor can be more compact and only deal with required properties.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
elefeint: This is catching the case where the result flux was empty, so the very first signal received is on complete. Without this check, isDmlQuery fails.
But it's a good point, I'll use the chaining.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
elefeint: As discussed, let's keep the minimal configuration (optional setter), and create a special-purpose StatementConfiguration object if the number of options grows unwieldy.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProviderTest.java
elefeint: It better be! I was stuck for two days writing it.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -102,8 +104,13 @@ public Batch createBatch() {
 
   @Override
   public Statement createStatement(String sql) {
-    return new SpannerStatement(this.client, this.session,
-        this.transactionContext, sql);
+    SpannerStatement statement
+        = new SpannerStatement(this.client, this.session, this.transactionContext, sql);
+
+    if (this.partialResultSetFetchSize != null) {
+      statement.setPartialResultSetPrefetch(this.partialResultSetFetchSize);
```
meltsufin: s/setPartialResultSetPrefetch/setPartialResultSetFetchSize for consistency

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -102,8 +104,13 @@ public Batch createBatch() {
 
   @Override
   public Statement createStatement(String sql) {
-    return new SpannerStatement(this.client, this.session,
-        this.transactionContext, sql);
+    SpannerStatement statement
+        = new SpannerStatement(this.client, this.session, this.transactionContext, sql);
+
+    if (this.partialResultSetFetchSize != null) {
```
meltsufin: If you made this field nullable, i.e. `Integer` everywhere, you wouldn't need this if statement.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -97,19 +106,35 @@ public Builder setCredentials(GoogleCredentials credentials) {
...
       }
 
-      return new SpannerConnectionConfiguration(
+      SpannerConnectionConfiguration configuration = new SpannerConnectionConfiguration(
           this.projectId,
           this.instanceName,
           this.databaseName,
           this.credentials);
+
+      if (this.partialResultSetFetchSize != null) {
```
meltsufin: Why the `if` statement here?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactory.java
```diff
@@ -38,9 +38,15 @@ public SpannerConnectionFactory(Client client, SpannerConnectionConfiguration co
   }
 
   @Override
-  public Publisher<SpannerConnection> create() {
+  public Mono<SpannerConnection> create() {
     return this.client.createSession(this.config.getFullyQualifiedDatabaseName())
-      .map(session -> new SpannerConnection(this.client, session));
+      .map(session -> {
+        SpannerConnection connection = new SpannerConnection(this.client, session);
+        if (this.config.getPartialResultSetFetchSize() != null) {
```
meltsufin: Again, I don't think the `if` is needed here.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -180,4 +186,13 @@ private boolean isSelectQuery() {
   private static boolean isDmlQuery(PartialResultSet firstPartialResultSet) {
     return firstPartialResultSet.getMetadata().getRowType().getFieldsList().isEmpty();
   }
+
+  /**
+   * Allows customizing the number of {@link PartialResultSet} objects to request at a time.
+   * @param prefetch batch size to request from Cloud Spanner
+   */
+  public void setPartialResultSetPrefetch(int prefetch) {
```
meltsufin: Need to rename this.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
dzou: I think this is unavoidable because this setting is propagated from the connectionFactory options

ConnectionFactory --> Connection --> Statement

Thus both Connection and Statement have the setting as a field to pass it downwards.

And I would not recommend the `Integer` approach because of confusing side effects.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
dzou: Can we just leave this `int`? Moving forward, I don't think we'll be remembering the reason it was `Integer` and it seems like an unusual practice.

dzou: Edit: nevermind

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
elefeint: going to make it integer everywhere except the getter, and add tests to catch null autoboxing

PR #97, opened 2019-06-05
TCK tests and fixes

### LINKED ISSUE #12: Bring in R2DBC SPI TCK
https://github.com/r2dbc/r2dbc-spi/tree/master/r2dbc-spi-test

ChengyuanZhao:
fixes #12 
makes some fixes as necessary


src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -88,7 +89,7 @@ public SpannerConnection(Client client, Session session) {
 
   @Override
   public Publisher<Void> close() {
-    return this.client.deleteSession(this.session);
+    return commitTransaction().concatWith(this.client.deleteSession(this.session));
```
meltsufin: What if there is no active transaction? This will cause a warning to be emitted.

dzou: I think using `.then` instead of `.concatWith` would be more accurate since `then` returns a Mono indicating completion; however concatWith returns a flux.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -98,16 +97,28 @@ public Statement add() {
 
   @Override
   public Statement bind(Object identifier, Object value) {
-    requireNonNull(identifier);
+    Assert.requireNonNull(identifier, "Identifier must not be null.");
```
dzou: Add another `Assert.requireNonNull(value)` here because value can no longer be null now right? It must be TypedNull?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -98,16 +97,28 @@ public Statement add() {
...
+        valToStore = null;
+        classToStore = ((TypedNull) value).getType();
+      } else {
+        valToStore = value;
+        classToStore = value.getClass();
+      }
+
       Codec codec = this.resolvedCodecs
-          .computeIfAbsent(paramName, n -> this.codecs.getCodec(value));
+          .computeIfAbsent(paramName, n -> this.codecs.getCodec(classToStore));
```
dzou: Should it be `this.codecos.getCodec(**valToStore**)` instead? I thought `this.codecs.getCodec` accepted the object value and not a Class.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -124,7 +135,7 @@ public Statement bind(int i, Object o) {
 
   @Override
   public Statement bindNull(Object identifier, Class<?> type) {
-    return bind(identifier, null);
+    return bind(identifier, new TypedNull(type));
```
dzou: You can just make this a singleton `private static final` instance to save yourself reinstantiating this.

meltsufin: In H2, they use `this.codecs.encodeNull(type)`. Can we do the same instead of creating this dummy `TypedNull` class?
@dmitry-s

src/main/java/com/google/cloud/spanner/r2dbc/codecs/DefaultCodecs.java
```diff
@@ -33,7 +34,7 @@
  */
 public final class DefaultCodecs implements Codecs {
 
-  private static final com.google.protobuf.Value NULL_VALUE =
+  public static final com.google.protobuf.Value NULL_VALUE =
```
dzou: Make this package-private?

src/main/java/com/google/cloud/spanner/r2dbc/codecs/Codecs.java
```diff
@@ -47,6 +47,6 @@
    */
   Value encode(Object value);
 
-  Codec getCodec(Object value);
```
meltsufin: This would  be a departure from other R2DBC implementations that all operate on the value rather than the type.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
ChengyuanZhao: This can't work because you need the codec of the right type even if the value is null.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
ChengyuanZhao: What do you mean??

ChengyuanZhao: I think that's arguably a WORSE way of doing it because the null must be bound to the statement with a type, which is why the individual codec must be found.

src/main/java/com/google/cloud/spanner/r2dbc/codecs/Codecs.java
ChengyuanZhao: This cannot work. A null must come with a type. Furthermore this is our interface and this is a big difference in Spanner vs old-fashioned sql servers that they've exclusively targeted so far.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
dzou: My bad, I thought u could just create a static instance of TypedNull and reuse that; however i forgot you had to pass the type into the constructor.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
ChengyuanZhao: will fix

src/main/java/com/google/cloud/spanner/r2dbc/codecs/DefaultCodecs.java
ChengyuanZhao: done

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
ChengyuanZhao: done

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
ChengyuanZhao: Fixed, I made it so that close() will not cause the warning to be logged.

I made close() do commit() because from the TCK tests it appears that using close() to commit a transaction (with no explicit call to commitTransaction) is an acceptable way to commit DML statements.

src/main/java/com/google/cloud/spanner/r2dbc/codecs/Codecs.java
ChengyuanZhao: https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/issues/101

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
ChengyuanZhao: https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/issues/101

ChengyuanZhao:
I'm gonna be moving on to the large integration tests issue next, so I will be submitting this in 15 minutes if no objections before then.

dzou:
> I'm gonna be moving on to the large integration tests issue next, so I will be submitting this in 15 minutes if no objections before then.

SGTM

PR #99, opened 2019-06-05
Add Helper Classes to Parse Statement Types

dzou:
This adds the initial helper classes to parse statement types.

Contributes to DML and DDL efforts in which statement type must be known.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
@@ -166,7 +166,7 @@ public class SpannerStatement implements Statement {
       return resultSetFlux
           .last()
           .map(partialResultSet -> Math.toIntExact(partialResultSet.getStats().getRowCountExact()))
-          .transform(rowCount -> Mono.just(new SpannerResult(Flux.empty(), rowCount)));
+          .map(rowCount -> new SpannerResult(Flux.empty(), Mono.just(rowCount)));
     }
   }
 
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
@@ -158,15 +158,15 @@ public class SpannerStatement implements Statement {
             this.session, this.transaction, this.sql, params, this.types);
 
     if (statementType == StatementType.SELECT) {
-      return Mono.just(new SpannerResult(
-          resultSetFlux.flatMapIterable(partialResultRowExtractor, getPartialResultSetFetchSize()),
-          Mono.just(0)));
+      return resultSetFlux
+          .flatMapIterable(partialResultRowExtractor, getPartialResultSetFetchSize())
+          .transform(result -> Mono.just(new SpannerResult(result, Mono.just(0))))
+          .next();
     } else {
-      Mono<Integer> rowsChanged =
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/statement/StatementParser.java
```diff
@@ -0,0 +1,65 @@
...
+      return StatementType.DDL;
+    } else if (isDml(processedSql)) {
+      return StatementType.DML;
+    } else {
+      return StatementType.UNKNOWN;
+    }
+  }
+
+  private static boolean isSelectQuery(String sql) {
+    return statementStartsWith(sql, SELECT_STATEMENTS);
```
dmitry-s: maybe just check in place without these one-line methods?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -140,53 +142,32 @@ public Statement bindNull(int i, Class<?> type) {
...
-        })
-        .next();
-  }
-
-  /**
-   * Returns whether we think the query is a DML query.
-   */
-  private static boolean isDmlQuery(PartialResultSet firstPartialResultSet) {
-    return firstPartialResultSet.getMetadata().getRowType().getFieldsList().isEmpty();
+    StatementType statementType = StatementParser.getStatementType(this.sql);
```
elefeint: Can doing it multiple times (in `execute` and `runSingleStatement()`, which gets run multiple times for each set of parameters) be avoided? Perhaps pass to `runSingleStatement()` in a parameter.

src/main/java/com/google/cloud/spanner/r2dbc/statement/StatementParser.java
```diff
@@ -0,0 +1,65 @@
...
+  private static final List<String> DDL_STATEMENTS = Arrays.asList("create", "drop", "alter");
+  private static final List<String> DML_STATEMENTS = Arrays.asList("insert", "update", "delete");
+
+  private StatementParser() {}
+
+  /**
+   * Returns the statement type of a given SQL string.
+   */
+  public static StatementType getStatementType(String sql) {
+    String processedSql = sql.trim().toLowerCase();
```
elefeint: I'm still iffy on lowercasing the whole string. Queries can get large.

src/main/java/com/google/cloud/spanner/r2dbc/statement/StatementParser.java
dzou: Done.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
dzou: Done.

src/main/java/com/google/cloud/spanner/r2dbc/statement/StatementParser.java
dzou: Yeah I agree. Sounds good, I can hold off on optimizing it for now; we can figure out a better method in the future.

dzou:
> We should probably take care of [ table_hint_expr ][ join_hint_expr ] as they can be present in the beginning of the query.
> See https://cloud.google.com/spanner/docs/query-syntax

Done.

PR #107, opened 2019-06-10
Batch DML support

### LINKED ISSUE #92: Natively support multiple DML statements
Cloud Spanner has several APIs for executing different, although somewhat overlapping, types of queries:
1) [ExecuteStreamingSql](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Spanner.ExecuteStreamingSql): supports insert/update statements in addition to the select statements (for which it is most useful). 
However, does NOT support DDL, and requires that DML statements arrive in order specified by `seqno`.
2) [ExecuteBatchDml](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Spanner.ExecuteBatchDml): supports multiple insert/update statements.
3) (out of scope for this issue) Database Admin API [UpdateDatabaseDdl](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.admin.database.v1#google.spanner.admin.database.v1.DatabaseAdmin.UpdateDatabaseDdl). This poses additional complications, as it returns a long-running operation, which then would have to be checked for status separately.

In #87, using multiple `ExecuteStreamingSql` calls when executing parameterized DML queries was made possible by explicitly serializing the DML queries to ensure they arrive to Spanner in order. However, waiting for the network roundtrip for each statement is not ideal. 
Since we will have to detect the nature of each query no matter what implementation we choose (because Spanner has separate APIs for different query types), we should use `ExecuteBatchDml` to execute multiple DML queries.

dmitry-s:
fixes #92

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
@@ -25,6 +25,7 @@ import com.google.cloud.spanner.r2dbc.statement.StatementParser;
 import com.google.cloud.spanner.r2dbc.statement.StatementType;
 import com.google.cloud.spanner.r2dbc.util.Assert;
 import com.google.protobuf.Struct;
+import com.google.spanner.v1.ExecuteBatchDmlResponse;
 import com.google.spanner.v1.PartialResultSet;
 import com.google.spanner.v1.Session;
 import com.google.spanner.v1.Type;
@@ -66,6 +67,8 @@ public class SpannerStatement implements Statement {
 
   private Map<String, Codec> resolvedCodecs = new HashMap<>();
 
+  private StatementType statementType;
+
   /**
    * Creates a Spanner statement for a given SQL statement.
    *
@@ -86,6 +89,7 @@ public class SpannerStatement implements Statement {
     this.session = session;
     this.transaction = transaction;
     this.sql = Assert.requireNonNull(sql, "SQL string can not be null");
+    this.statementType = StatementParser.getStatementType(this.sql);
   }
 
   @Override
@@ -154,23 +158,26 @@ public class SpannerStatement implements Statement {
     }
 
     Flux<Struct> structFlux = Flux.fromIterable(this.bindingsStucts);
-    StatementType statementType = StatementParser.getStatementType(this.sql);
 
-    if (statementType == StatementType.SELECT) {
-      return structFlux.flatMap(struct -> runSingleStatement(struct, statementType));
+    if (this.statementType == StatementType.DML) {
+      return this.client
+          .executeBatchDml(this.session, this.transaction, this.sql, this.bindingsStucts,
+              this.types)
+          .flatMapIterable(ExecuteBatchDmlResponse::getResultSetsList)
+          .map(resultSet -> new SpannerResult(Flux.empty(),
+              Mono.just(Math.toIntExact(resultSet.getStats().getRowCountExact()))));
     }
-    // DML statements have to be executed sequentially because they need seqNo to be in order
-    return structFlux.concatMapDelayError(struct -> runSingleStatement(struct, statementType));
+    return structFlux.flatMap(this::runSingleStatement);
   }
 
-  private Mono<? extends Result> runSingleStatement(Struct params, StatementType statementType) {
+  private Mono<? extends Result> runSingleStatement(Struct params) {
     PartialResultRowExtractor partialResultRowExtractor = new PartialResultRowExtractor();
 
     Flux<PartialResultSet> resultSetFlux =
         this.client.executeStreamingSql(
             this.session, this.transaction, this.sql, params, this.types);
 
-    if (statementType == StatementType.SELECT) {
+    if (this.statementType == StatementType.SELECT) {
       return resultSetFlux
           .flatMapIterable(partialResultRowExtractor, getPartialResultSetFetchSize())
           .transform(result -> Mono.just(new SpannerResult(result, Mono.just(0))))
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -108,24 +112,27 @@ public Statement bindNull(int i, Class<?> type) {
...
-  private Mono<? extends Result> runSingleStatement(Struct params, StatementType statementType) {
+  private Mono<? extends Result> runSingleStatement(Struct params) {
     PartialResultRowExtractor partialResultRowExtractor = new PartialResultRowExtractor();
 
     Flux<PartialResultSet> resultSetFlux =
         this.client.executeStreamingSql(
             this.session, this.transaction, this.sql, params, this.statementBindings.getTypes());
 
-    if (statementType == StatementType.SELECT) {
+    if (this.statementType == StatementType.SELECT) {
```
elefeint: Do we need this check anymore? DML does not go through this method anymore, and I don't imagine DDL will either.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -159,6 +162,31 @@ public GrpcClient(GoogleCredentials credentials) {
...
+    }
+    for (Struct paramsStruct : params) {
+      ExecuteBatchDmlRequest.Statement statement = ExecuteBatchDmlRequest.Statement.newBuilder()
+          .setSql(sql).setParams(paramsStruct).putAllParamTypes(types)
+          .build();
+      request.addStatements(statement);
+    }
+
+    return ObservableReactiveUtil
+        .streamingCall(obs -> this.spanner.executeBatchDml(request.build(), obs));
```
elefeint: It's not really a streaming call; use `unaryCall()` here.

src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerIT.java
```diff
@@ -268,12 +292,13 @@ public void testEmptySelect() {
   /**
    * Executes a DML query and returns the rows updated.
    */
-  private int executeDmlQuery(String sql) {
+  private List<Integer> executeDmlQuery(String sql) {
```
dzou: Should this be kept as before, returning integer? B/c `String sql` can only correspond to one statement.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -108,24 +112,27 @@ public Statement bindNull(int i, Class<?> type) {
 
   @Override
   public Publisher<? extends Result> execute() {
-    Flux<Struct> structFlux = Flux.fromIterable(this.statementBindings.getBindings());
-    StatementType statementType = StatementParser.getStatementType(this.sql);
-
-    if (statementType == StatementType.SELECT) {
-      return structFlux.flatMap(struct -> runSingleStatement(struct, statementType));
+    if (this.statementType == StatementType.DML) {
```
meltsufin: I wonder if you should just have a class hierarchy of statements -- `SpannerDmlStatement`, `SpannerDdlStatement`, `SpannerReadStatement`. WDYT?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -108,24 +112,27 @@ public Statement bindNull(int i, Class<?> type) {
...
+          .executeBatchDml(this.session, this.transaction, this.sql,
+              this.statementBindings.getBindings(),
+              this.statementBindings.getTypes())
+          .flatMapIterable(ExecuteBatchDmlResponse::getResultSetsList)
+          .map(resultSet -> new SpannerResult(Flux.empty(),
+              Mono.just(Math.toIntExact(resultSet.getStats().getRowCountExact()))));
     }
-    // DML statements have to be executed sequentially because they need seqNo to be in order
-    return structFlux.concatMapDelayError(struct -> runSingleStatement(struct, statementType));
+    Flux<Struct> structFlux = Flux.fromIterable(this.statementBindings.getBindings());
```
meltsufin: Else if `SELECT`? Are you missing the DDL case?
The whole thing cloud be a switch statement, with the default throwing an exception.
Alternatively, see my suggestion about the class hierarchy.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
meltsufin: +1 and should probably rename to `runSelectStatement`

src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
```diff
@@ -83,6 +85,13 @@
     return  executeStreamingSql(session, transaction, sql, null, null);
   }
 
+  /**
+   * Execute DML batch.
+   */
+  Flux<ExecuteBatchDmlResponse> executeBatchDml(Session session,
+      @Nullable SpannerTransactionContext transactionContext, String sql,
+      List<Struct> params, Map<String, Type> types);
```
meltsufin: Should `params` be a `Flux<Struct>`, or is that a higher level concern?

src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerIT.java
```diff
@@ -268,12 +292,13 @@ public void testEmptySelect() {
...
     Connection connection = Mono.from(connectionFactory.create()).block();
 
     Mono.from(connection.beginTransaction()).block();
-    int rowsUpdated = Mono.from(connection.createStatement(sql).execute())
+    List<Integer> rowsUpdated = Flux.from(connection.createStatement(sql).execute())
```
meltsufin: `s/rowsUpdated/rowsUpdatedPerStatement`

src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
dmitry-s: Yes, it is a higher level concern.

Issue #40, opened 2019-05-10
Extend SpannerConnection to support multiple transaction types


dzou:
Currently `SpannerConnection` defaults to using read-write transactions.

Extend it to support `read-only` and `partitioned DML` transactions.

elefeint:
NB: partitioned DML transactions don't need to be committed (don't support commit or rollback).

dzou:
From the master design doc:

> The SPI supports isolation level specification on the Connection, but Cloud Spanner just has two types of transactions read-only and read-write, which don’t map into the isolation levels.
The decision is to allow configuring the default transaction mutability on the ConnectionFactory as well as overloading the beginTrasaction(mutability) method in the SpannerConnection class. In the future, the R2DBC SPI may support the transaction mutability setting natively.

dzou:
This one is a little bit unusual; on further experimentation, it appears that it is invalid to call `CommitTransaction` on a read-only or partitioned DML transaction. Currently I'm thinking of just doing a no-op commit if user attempts to call commit in either context.

ChengyuanZhao:
looks like this was completed in #109

PR #109, opened 2019-06-11
Add Support for Different Transaction Types

dzou:
Adds an additional overloaded method `beginTransaction(transactionType)` to `SpannerConnection` in order to support multiple transaction types: read-write, read-only, and partitioned DML.

Notes:
- Commit Transaction is not a valid operation for read-only and partitioned DML transactions. Thus, commit will be a no-op if called during one of these transaction types.
- For partitioned dml statements, exact row count updated is not returned but a lower bound is returned instead. Thus this number is used to populate SpannerResult.

In reference to: #40.


src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -89,15 +88,14 @@ public GrpcClient(GoogleCredentials credentials) {
...
+
     return Mono.defer(() -> {
       BeginTransactionRequest beginTransactionRequest =
           BeginTransactionRequest.newBuilder()
               .setSession(session.getName())
-              .setOptions(
-                  TransactionOptions
-                      .newBuilder()
-                      .setReadWrite(ReadWrite.getDefaultInstance()))
+              .setOptions(transactionOptions)
```
ChengyuanZhao: In addition to setting the transaction options here, don't you also need to actually swap out the call for from executeStreamingSql to partitionQuery for example?

dzou: Hmm am I mistaken about the Partitioned DML - but that is executed through executeStreamingSQL and not partitionQuery right?

see: https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#partitionqueryrequest
> This must not contain DML commands, such as INSERT, UPDATE, or DELETE. Use ExecuteStreamingSql with a PartitionedDml transaction for large, partition-friendly DML operations.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -31,6 +33,11 @@
  */
 public class SpannerConnection implements Connection {
 
+  private static final TransactionOptions READ_WRITE_TRANSACTION =
```
meltsufin: Does it make sense to just create our own `TransactionType` enum and not expose this gRPC level object to the clients?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -53,35 +60,42 @@ public SpannerConnection(Client client, Session session) {
   }
 
   @Override
-  public Publisher<Void> beginTransaction() {
-    return this.client.beginTransaction(this.session)
-        .doOnNext(
-            transaction -> this.transactionContext = SpannerTransactionContext.from(transaction))
+  public Mono<Void> beginTransaction() {
+    return this.beginTransaction(READ_WRITE_TRANSACTION);
```
meltsufin: Should the default come from the `ConnectionOptions` or something?

elefeint: This could be in scope for #119.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -53,35 +60,42 @@ public SpannerConnection(Client client, Session session) {
...
-      } else {
+      if (this.transactionContext != null && this.transactionContext.isReadWrite()) {
         return this.client.commitTransaction(this.session, this.transactionContext.getTransaction())
             .doOnNext(response -> this.transactionContext = null)
             .then();
       }
+
+      if (this.transactionContext == null) {
+        this.logger.debug("commitTransaction() is a no-op; called with no transaction active.");
+      } else if (this.transactionContext.isReadWrite()) {
```
elefeint: Should this be negated? The debug message says we are *not* in read-write here.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
dzou: Sounds good. For this one I realized that the READ-ONLY `transactionType` actually has a number of different options that the user can set: https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.TransactionOptions.ReadOnly

@elefeint, @meltsufin What are your thoughts on exposing these settings for read-only? Should we make this into a TransactionType class then with builder pattern?

meltsufin: In that case I would just expose `TransactionOptions` like you have it now. Ignore my enum suggestion.

src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerIT.java
```diff
@@ -80,6 +83,21 @@
           .option(DATABASE, TEST_DATABASE)
           .build());
 
+  private static final TransactionOptions READ_WRITE_TRANSACTION =
+      TransactionOptions.newBuilder()
+          .setReadWrite(ReadWrite.getDefaultInstance())
+          .build();
+
+  private static final TransactionOptions PARTITIONED_DML =
```
ChengyuanZhao: There might be some messiness when merging w/ Dmitry's batch code, because there we check for the query string's starting word and use the batch execute method for all DML. That might clash with the partitioned DML request type you have here.

dzou: Yeah I ran into that just now. Seems like one cannot use the BatchDDL method to run partitioned DML statements.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
dzou: Ok, leaving as is for now.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
dzou: Yeah for this PR I will only focus on adding a hook for setting transaction through `beginTransaction`; the connection options propagation will be deferred to a followup PR.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
dzou: Nice catch, done.

dzou:
> Looks good; `SpannerConnectionTest` needs new tests for corresponding functionality changes, and there may be a negation needed in an edge case.

Done. Added a test to `SpannerConnectionTest` to verify that the transaction options are propagated and also commit is a no-op for non-readWrite transactions. Also added an integration test with multiple transaction types.

PR #111, opened 2019-06-11
group session and transaction as a unified ExecutionContext

### LINKED ISSUE #98: Storing Transaction instead of a Mono of Transaction causes some issues
see #97 

One of the TCK tests ( "tranactionCommit()" ) begins a transaction and then follows it up with thenMany, but it seems the following operations try to use the transaction when it isn't set yet.

elefeint:
Turned `TransactionContext` into `ExecutionContext` by adding session information. In the object, session is required upon context creation, while transaction is optional. This solves the issue of session/transaction not yet existing when `SpannerStatement` gets created outside of publisher flow. 
 
It looks to be a convenient abstraction even on its own, since nearly all operations require a session (except for creating a session, natch), and many operations are transaction-aware.

Also:
* Added an assert to guard against String/proto existential emptiness. 
* Fixed transaction not being unset after rollback.
* Made `executeBatchDml` payload deferred to avoid reintroducing the issue with null session and transaction being injected into the request.

Fixes #98.


src/main/java/com/google/cloud/spanner/r2dbc/ExecutionContext.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package com.google.cloud.spanner.r2dbc;
+
+import com.google.protobuf.ByteString;
+
+/**
+ * Contract for objects encapsulating Spanner state (session, transaction, update sequence).
+ */
+public interface ExecutionContext {
```
dzou: Would it be better to just use a concrete class `SpannerExecutionContext` and omit this interface? I think this approach may be simpler since we do not intend to expose `ExecutionContext` to the user.

elefeint: I wanted to keep the implementation private while still leaving a mockable interface. End users can also get a readonly context instance from the `Connection`; it would be less convenient to have to use `SpannerConnection.ExecutionContext`. On the other hand, if I pull up the implementation to become a standalone class, I can't keep the transaction setter private.

meltsufin: The name is a bit too abstract IMO. `SpannerStatementExecutionContext`?

meltsufin: Why do you need it to be mockable? This is really just a struct. It has no behavior.

src/main/java/com/google/cloud/spanner/r2dbc/ExecutionContext.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner.r2dbc;
+
+import com.google.protobuf.ByteString;
+
+/**
+ * Contract for objects encapsulating Spanner state (session, transaction, update sequence).
```
dzou: Can you add a little bit more elaboration for the javadoc; indicating these properties are relevant to the Spanner connection?

elefeint: yep!

src/main/java/com/google/cloud/spanner/r2dbc/ExecutionContext.java
elefeint: Mockable is not necessary but is nice to have because the actual implementation has sequence-incrementing behavior.

A more pressing concern for me was restricting transaction modification to only be possible from `SpannerConnection` while still having public (I could settle for default/package-private, but public seems harmless-to-potentially-useful)  getter access.

dzou: Isn't `SpannerExecutionContext` already immutable though? i.e. it appears to only have getters exposed

elefeint: Right, `setTransaction()` is private, so only its parent class `SpannerConnection` can call it.
If I remove the interface, then callers will have to reference `SpannerConnection.SpannerExecutionContext`, which seems clunky.
If I replace the interface with the concrete class, I won't be able to keep `setTransaction()` private.

dzou: Gotcha, I see.

src/main/java/com/google/cloud/spanner/r2dbc/client/Client.java
```diff
@@ -44,52 +43,48 @@
...
+  Mono<Void> deleteSession(ExecutionContext ctx);
 
   /**
    * Begins a new Spanner {@link Transaction} within the provided {@link Session}.
-   * @param session The {@link Session} object with which requests are made to the Spanner API.
+   * @param ctx Execution context containing the current session.
    * @returns {@link Mono} of the transaction that was started.
    */
-  Mono<Transaction> beginTransaction(Session session);
+  Mono<Transaction> beginTransaction(ExecutionContext ctx);
```
ChengyuanZhao: I don't think it's appropriate to require beginTransaction take an entire context instead of just the session. 

The other two items in the context (the transaction id and the seq no) both start fresh for a new transaction, so it is awkward they're also provided to beginTransaction.

elefeint: yeah, I agree. I'll change that one back. I did it for consistency and because I was on a roll of copy/paste.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProviderTest.java
```diff
@@ -62,13 +62,17 @@
 
   SpannerConnectionFactoryProvider spannerConnectionFactoryProvider;
 
-  Client mockClient = Mockito.mock(Client.class);
+  Client mockClient;
+
+  StatementExecutionContext mockContext;
```
dzou: May remove this as it does not appear to be used by this test.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -42,11 +46,9 @@
 
   private final Client client;
 
-  private final Session session;
-
   private final SpannerConnectionConfiguration config;
 
-  private volatile SpannerTransactionContext transactionContext;
+  private volatile SpannerStatementExecutionContext ctx;
```
ChengyuanZhao: Should this be `private final` since it's always there but we set the transaction inside of it?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -159,10 +155,74 @@ public SpannerStatement createStatement(String sql) {
...
+   * A class to hold session and transaction-related data that needs to be communicated from
+   * {@link SpannerConnection} to {@link SpannerStatement}.
    */
-  public Session getSession() {
-    return this.session;
+  public static class SpannerStatementExecutionContext implements StatementExecutionContext {
+
+    private Session session;
+
+    private Transaction transaction;
```
ChengyuanZhao: Is this now the `volatile` item?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -159,10 +155,74 @@ public SpannerStatement createStatement(String sql) {
...
+    public long nextSeqNum() {
+      return this.seqNum.getAndIncrement();
+    }
+
+    public boolean isTransactionReadWrite() {
+      return this.transactionOptions.hasReadWrite();
+    }
+
+    public boolean isTransactionPartitionedDml() {
+      return this.transactionOptions.hasPartitionedDml();
```
ChengyuanZhao: I suppose we wouldn't run into an NPE here if we don't make mistakes elsewhere, but might be convenient to return false is transactionOptions is null. (and the one above ? )

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProviderTest.java
elefeint: It's there to prevent `SpannerConnectionFactoryProvider` from instantiating a real client (and creating a gRPC channel). I'll move it inline, though, it does not need to be a class member.

elefeint: Never mind, it's used, just not in scope for my changes. I moved it to `setUp()` for consistency with context mock.

dzou: Ah this was referring to `this.mockContext = mock(StatementExecutionContext.class);`, I don't believe the mockContext is being used in the test; the client is being used though.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
elefeint: yeah, and you are right; this is no longer volatile.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProviderTest.java
elefeint: Right, I see it now. Removing.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
elefeint: I am going to skip adding volatile, since we did it for performance, and creating new transactions is not a high-volume operation.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
elefeint: Good point, I'll add a gating check.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -55,45 +57,43 @@
    */
   public SpannerConnection(Client client, Session session, SpannerConnectionConfiguration config) {
     this.client = client;
-    this.session = session;
+    this.ctx = new SpannerStatementExecutionContext(session);
```
meltsufin: Why not accept `SpannerStatementExecutionContext` in the constructor instead of `Session` then?
It would make everything more testable.

src/main/java/com/google/cloud/spanner/r2dbc/StatementExecutionContext.java
```diff
@@ -0,0 +1,42 @@
...
+ *   <li>Session is created per-connection and cannot be changed.
+ *   <li>Transaction is initially {@code null}, and can change as connection begins and ends
+ *   transactions.
+ *   <li>monotonically increasing update sequence describes the order of DML statements within a
+ *   transaction.
+ * </ul>
+ */
+public interface StatementExecutionContext {
+
+  public String getSessionName();
```
meltsufin: Typically for interfaces, methods should hava javadoc too.

src/main/java/com/google/cloud/spanner/r2dbc/util/Assert.java
```diff
@@ -45,4 +46,40 @@ private Assert() {}
...
+   * Checks that a specified object reference is not {@code null} or an empty string, and throws a
+   * customized {@link IllegalArgumentException} if it is.
+   *
+   * @param s string to check
+   * @param message informative message to be used in the event that an
+   * {@link IllegalArgumentException} is thrown
+   * @return the original string {@code s}
+   * @throws IllegalArgumentException if {@code o} is {@code null}
+   */
+  public static String requireNonEmpty(@Nullable String s, String message) {
```
meltsufin: Instead of repeating the same method, can it just be type paramaterized (generic)?

elefeint: StringValue is a protobuf that happens to have a method named `isEmpty()`. It does not share any inheritance hierarchy with String. 
If this were a language with prototypical inheritance *cough* javascript *cough*, it would be trivial.

In Java, short of doing reflection, I don't think it can be done.

ChengyuanZhao: Yea this is because while both types have a "isEmpty" method they aren't the same method. 

If this were a language that had ad-hoc polymorphism we could've made a type class called PossiblyEmpty or something and then provided implementations for both types then we could use a type constraint that the type variable is a PossiblyEmpty. Alas this is java :/

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
elefeint: `SpannerConnection` modifies the transaction within context; it might as well create it, too. I was trying to keep the context only modifiable from one class -- `SpannerConnection`. Everyone else gets read-only rights.

meltsufin: Can you try to just have `SpannerConnection` implement `SpannerStatementExecutionContext`?

src/main/java/com/google/cloud/spanner/r2dbc/util/Assert.java
meltsufin: That sucks. Ok then.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
elefeint: This was a good change -- it enabled the `isTransaction[PartitionedDml|ReadWrite]` to become private.

Issue #93, opened 2019-06-04
Sample+README and Main Ref Doc 


ChengyuanZhao:

meltsufin:
The main ref doc should just go into the README for now.

ChengyuanZhao:
Closing because SAMPLE is done , and we have individual readme section issues separately.

PR #120, opened 2019-06-12
DDL DML SQL sample and readme

ChengyuanZhao:
related #93 
sample with DML, DDL, SQL and readme

### DIFF
--- a/sample/src/main/java/com/example/BookExampleApp.java
+++ b/sample/src/main/java/com/example/BookExampleApp.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


sample/src/main/java/com/example/BookExampleApp.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  public void createTable() {
+    Mono.from(connectionFactory.create())
+        .delayUntil(c -> c.createStatement("CREATE TABLE BOOKS ("
+            + "  ID STRING(20) NOT NULL,"
+            + "  TITLE STRING(MAX) NOT NULL"
+            + ") PRIMARY KEY (ID)").execute())
+        .doOnNext(x -> System.out.println("Table creation completed."))
+        .block();
```
dzou: Would it be better to write this sample not using blocks to better follow reactive patterns?

ChengyuanZhao: Regarding the spring thing: 
If we do that it must be a separate sample. There must be a sample about using the driver by itself, since this is the r2dbc module.

regarding chaining them:
I don't think so. I don't think a user would want to chain create and insert and delete together. Furthermore, I do show chaining with the doOnNext print calls.

elefeint: Maybe not spring in this example, but using [r2dbc-client ](https://github.com/r2dbc/r2dbc-client) may make sense.

Issue #113, opened 2019-06-11
Documentation: Select vs DML vs DDL


elefeint:
Document standard R2DBC treatment, explain that the driver will select the correct API to use.

Document the special `SpannerStatement` functionality that deviates from unified R2DBC treatment, when available.

dzou:
@elefeint - Could you elaborate on this part:

> Document the special SpannerStatement functionality that deviates from unified R2DBC treatment, when available.

I forgot what this referred to.

elefeint:
On statement:
* Lack of support for index binding
* explanation that the driver will pick one of 3 APIs (executeStreamingSql, batch update, DDL) depending on query content.

On connection:
* the custom transaction type override 
* lack of support for checkpoints.

PR #126, opened 2019-06-13
param binding readme

### LINKED ISSUE #118: Documentation: parameter binding
Discuss index-based column access support (or lack thereof).

ChengyuanZhao:
fixes #118 
param binding readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,8 @@ mySpannerConnection.createStatement(
     .add()
     .bind("id", "book-id-2")
     .bind("title", "Book Two")
-    .execute();
+    .execute()
+    .flatMap(r -> r.getRowsUpdated());
 ``` 
 
 The parameter identifiers must be `String`. 
--- a/README.md
+++ b/README.md
@@ -32,19 +32,22 @@ SQL and DML statements can be constructed with parameters:
 ```java
 mySpannerConnection.createStatement(
   "INSERT BOOKS (ID, TITLE) VALUES (@id, @title)")
-  .bind("id", "book-id-1")
-  .bind("title", "Book One")
-  .add()
-  .bind("id", "book-id-2")
-  .bind("title", "Book Two")
-  .execute();
+    .bind("id", "book-id-1")
+    .bind("title", "Book One")
+    .add()
+    .bind("id", "book-id-2")
+    .bind("title", "Book Two")
+    .execute();
 ``` 
 
 The parameter identifiers must be `String`. 
 Positional parameters are not supported.
 
 The example above binds two sets of parameters to a single DML template. 
-It will produce a `Flux` containing two `SpannerResult` objects for the two instances of the statement that are executed. 
+It will produce a `Publisher` (implemented by a `Flux`) containing two `SpannerResult` objects for the two instances of the statement that are executed. 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,6 +24,28 @@ Null values mapping is supported in both directions.
...
+  .bind("id", "book-id-2")
+  .bind("title", "Book Two")
+  .execute();
+``` 
+
+The parameter identifiers must be `String`. 
+Positional parameters are not supported.
+
+The example above binds two sets of parameters to a single DML template. 
+It will produce a `Flux` containing two `SpannerResult` objects for the two instances of the statement that are executed.
```
meltsufin: Maybe also show how to extract the results after executing the query?

README.md
```diff
@@ -24,6 +24,28 @@ Null values mapping is supported in both directions.
...
 See [Cloud Spanner documentation](https://cloud.google.com/spanner/docs/data-types) to learn more about Spanner types.
 
+## Binding Query Parameters
+
+The Cloud Spanner R2DBC driver supports named parameter binding using Cloud Spanner's [parameter syntax](https://cloud.google.com/spanner/docs/sql-best-practices).
+
+SQL and DML statements can be constructed with parameters:
+```java
+mySpannerConnection.createStatement(
+  "INSERT BOOKS (ID, TITLE) VALUES (@id, @title)")
```
dzou: nit: can you make the content that go to the new line have indentation level of 4 instead of 2?

README.md
dzou: Since `Project-Reactor` is supposed to be an implementation detail, maybe say `Publisher` instead of `Flux` instead?

ChengyuanZhao: I was thinking how to use `SpannerStatement` would be covered in this issue: https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/issues/113

That issue talks about DML vs DDL vs SQL which naturally would mention the `map()` and `getRowsUpdated` methods.

meltsufin: Does `execute()` actually execute the statement, or do you still need to block on the result?
Maybe clarify that?

ChengyuanZhao: I will add a sentence.

README.md
```diff
@@ -32,19 +32,22 @@ SQL and DML statements can be constructed with parameters:
...
 
 The parameter identifiers must be `String`. 
 Positional parameters are not supported.
 
 The example above binds two sets of parameters to a single DML template. 
-It will produce a `Flux` containing two `SpannerResult` objects for the two instances of the statement that are executed. 
+It will produce a `Publisher` (implemented by a `Flux`) containing two `SpannerResult` objects for the two instances of the statement that are executed. 
+
+Note that calling `execute` produces R2DBC `Result` objects, but this doesn't cause the query to be run on the database. 
+You must use the `map` or `getRowsUpdated` methods of the results to complete the underlying queries.
```
meltsufin: I think this should be part of the example. Otherwise we risk people copy-pasting this code and wondering why the query is not executing.

ChengyuanZhao: Done.

Though I must respectfully disagree with it. It's a slippery slope because this is R2DBC behavior, not R2DBC Cloud Spanner specific behavior.

PR #131, opened 2019-06-14
R2DBC batch support

### LINKED ISSUE #122: Support R2DBC Batch
See https://r2dbc.io/spec/0.8.0.M8/spec/html/#batches
Example:
```
Batch batch = connection.createBatch();
Publisher<? extends Result> publisher = batch.add("SELECT title, author FROM books")
    .add("INSERT INTO books VALUES('John Doe', 'HappyBooks LLC')
    .execute();
```

dmitry-s:
fixes #122


src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
```diff
@@ -0,0 +1,70 @@
...
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+/**
+ * {@link Batch} implementation for Cloud Spanner.
+ */
+public class SpannerBatch implements Batch {
+
+  private SpannerTransactionContext transactionContext;
```
dzou: Will have to merge and move to using `StatementExecutionContext` after Elena's changes.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
```diff
@@ -0,0 +1,70 @@
...
+  private Session session;
+  private Client client;
+
+  List<String> statements = new ArrayList<>();
```
elefeint: private

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
```diff
@@ -0,0 +1,70 @@
...
+    this.statements.add(sql);
+    return this;
+  }
+
+  @Override
+  public Publisher<? extends Result> execute() {
+    return this.client
+        .executeBatchDml(this.session, this.transactionContext, this.statements)
+        .flatMapIterable(ExecuteBatchDmlResponse::getResultSetsList)
+        .map(resultSet -> Math.toIntExact(resultSet.getStats().getRowCountExact()))
```
elefeint: Might as well collapse into a single map that does both.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -183,7 +183,37 @@ public GrpcClient(GoogleCredentials credentials) {
   public Mono<ExecuteBatchDmlResponse> executeBatchDml(Session session,
       @Nullable SpannerTransactionContext transactionContext, String sql,
       List<Struct> params, Map<String, Type> types) {
+    ExecuteBatchDmlRequest.Builder request = createBatchDmlRequestBuilderBuilder(session,
```
elefeint: LOL for `BuilderBuilder`. I think `create...Builder` with a single Builder already gets the point across though.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
```diff
@@ -0,0 +1,78 @@
...
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+/**
+ * {@link Batch} implementation for Cloud Spanner.
+ */
+public class SpannerBatch implements Batch {
+
+  private final SpannerConnection connection;
```
elefeint: `SpannerConnection` implements `StatementExecutionContext`, so you don't need `ctx` anymore. In fact, now that statements and batches have a reference back to `SpannerConnection`, I wonder if I should remove the context interface completely.

On the other hand, having the mockable interface makes statements/batches easier to test -- with the above, you'd have to instantiate a `SpannerConnection` and start a transaction to have a proper setup for a "create batch assuming that a transaction is already present" scenario.

@meltsufin What do you think?

src/test/java/com/google/cloud/spanner/r2dbc/SpannerStatementTest.java
```diff
@@ -56,8 +58,8 @@
  */
 public class SpannerStatementTest {
 
-  private static final Session TEST_SESSION =
-      Session.newBuilder().setName("project/session/1234").build();
+  @Rule
```
elefeint: The other tests use `assertThatThrownBy()`; could you switch to that for consistency?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
dzou: Hmm, since both statement and batch require a reference back to a connection now, I think I would be in favor for removing the interface. It would remove one level of indirection for us and simplify things.

I think the testing benefits from the mockable interface are not as strong because one is still be able to mock a SpannerConnection using Mockito, so you won't have to go through the hoops of building a SpannerConnection if you don't want to.

meltsufin: I don't really understand why we need the `SpannerConnection` here, but not in `SpannerStatement`.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
```diff
@@ -0,0 +1,78 @@
...
+    return executeInTransaction().flatMapIterable(ExecuteBatchDmlResponse::getResultSetsList)
+        .map(resultSet -> {
+          int count = Math.toIntExact(resultSet.getStats().getRowCountExact());
+          return new SpannerResult(Flux.empty(), Mono.just(count));
+        });
+  }
+
+  private Mono<ExecuteBatchDmlResponse> executeInTransaction() {
+    return this.ctx.getTransactionId() == null
+        ? this.connection.beginTransaction()
```
meltsufin: Shouldn't the client manage the transaction? Why are we starting it here? This goes back to my question about why we need the connection in `SpannerBatch`?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
dzou: O we need it in `AutoCommitSpannerStatement` - this was going back to our discussion about whether we believed a subclass of SpannerStatement was necessary. If we omit this subclass, then SpannerStatement will need the SpannerConnection.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
dmitry-s: that's exactly how we handle transactions in `AutoCommitSpannerStatement`, and we have a `SpannerConnection` reference there

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+public class SpannerBatch implements Batch {
+
+  private final SpannerConnection connection;
+  private final StatementExecutionContext ctx;
+  private final Client client;
+
+  private List<String> statements = new ArrayList<>();
+
+  SpannerBatch(Client client, StatementExecutionContext ctx, SpannerConnection connection) {
```
dzou: Because SpannerConnection implements StatementExecutionContext, it is sufficient to only include the SpannerConnection in the constructor.

dmitry-s: It looks like `StatementExecutionContext` should not exist at all

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
meltsufin: So batch statements should always be in auto-commit mode? Is that defined in the spec?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
dmitry-s: But that's probably beyond the scope of this PR

meltsufin: I think it makes sense in `SpannerStatement`.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
meltsufin: I don't think the R2DBC spec assumes that the statement even knows about the connection. The transactions should be managed by the connection, not the statement. The auto-commit is an interesting case, and maybe we can rethink how we do it to avoid giving the connection to the statement. Maybe it's the `Client` that needs to start the transaction for auto-commit.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerBatch.java
dmitry-s: Spec doesn't mention transactions. But it looks like it is run in an auto-commit transaction in the example they used.

meltsufin: Can we add auto-commit as a flag into the `StatementExecutionContext` and then have the `Client` use it?

PR #133, opened 2019-06-17
Single use autocommit dml statement

### LINKED ISSUE #119: Investigate autocommit for DML
R2DBC's default behavior is autocommit.
Cloud Spanner's natural default behavior is a readonly transaction.

But @dzou had an idea that it is possible for us to support autocommit for autodetected DML statement type.

ChengyuanZhao:
fixes #119 
Single use autocommit dml statement


src/main/java/com/google/cloud/spanner/r2dbc/IndependentSpannerStatement.java
```diff
@@ -0,0 +1,58 @@
...
+ * A `SpannerStatement` that is single-use for read and write.
+ */
+public class IndependentSpannerStatement extends SpannerStatement {
+
+  private final SpannerConnection spannerConnection;
+
+  /**
+   * Creates a Spanner statement for a given SQL statement.
+   *
+   * <p>Session and transaction may not be present at the time this statement is created, therefore
```
elefeint: Remove this paragraph -- it does not apply because you don't take session/transaction as parameters, only the spanner connection.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -138,10 +138,9 @@ public Batch createBatch() {
 
   @Override
   public SpannerStatement createStatement(String sql) {
-    SpannerStatement statement
-        = new SpannerStatement(this.client, this, sql, this.config);
-
-    return statement;
+    return getTransactionId() == null
```
elefeint: creating/committing a transaction for all queries will slow down readonly queries. Would it make sense to check `sql` here and only use `IndependentSpannerStatement` for DML queries?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
```diff
@@ -40,7 +40,7 @@
 
   private Client client;
 
-  private StatementExecutionContext ctx;
+  protected StatementExecutionContext ctx;
```
elefeint: `IndependentSpannerStatement` does not use context, does it? And it has access to `SpannerConnection`, which implements the context.

src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerIT.java
```diff
@@ -229,6 +229,59 @@ public void testSessionManagement() {
...
+                    .bind("category", 100L)
+                    .bind("title", "b3")
+                    .bind("fiction", false)
+                    .bind("published", LocalDate.of(2016, 1, 6))
+                    .bind("wps", 15.22)
+                    .execute())
+                    .flatMapSequential(r -> Mono.from(r.getRowsUpdated())))
+                    .expectNext(1).expectNext(1).expectNext(1).verifyComplete())
+        )
+        .block();
```
elefeint: StepVerifier?

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
```diff
@@ -148,6 +148,18 @@ public void executeSingleUseDml() {
     assertThat(statement).isInstanceOf(IndependentSpannerStatement.class);
   }
 
+  @Test
+  public void executeDmlInTransactionTest() {
+    SpannerConnection connection
+        = new SpannerConnection(this.mockClient, TEST_SESSION, TEST_CONFIG);
+    String sql = "insert into books values (title) @title";
+
+    connection.beginTransaction().block();
```
elefeint: StepVerifier that chains transaction with statement and then verifies the `SpannerStatement` type within the step verifier flow?

src/main/java/com/google/cloud/spanner/r2dbc/IndependentSpannerStatement.java
ChengyuanZhao: done

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
ChengyuanZhao: totally right! fixed. covered with test!

src/main/java/com/google/cloud/spanner/r2dbc/SpannerStatement.java
ChengyuanZhao: fixed.

src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerIT.java
ChengyuanZhao: done

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
ChengyuanZhao: done

src/main/java/com/google/cloud/spanner/r2dbc/IndependentSpannerStatement.java
```diff
@@ -0,0 +1,55 @@
...
+   */
+  public IndependentSpannerStatement(Client client, String sql,
+      SpannerConnectionConfiguration config, SpannerConnection spannerConnection) {
+    super(client, spannerConnection, sql, config);
+    this.spannerConnection = Assert
+        .requireNonNull(spannerConnection, "A non-null SpannerConnection is required.");
+  }
+
+  @Override
+  public Publisher<? extends Result> execute() {
```
dzou: Would this become problematic if the user creates a statement like this and doesn't run it, and then executes it after they begin a transaction?

i.e.

```
Statement s = connection.createStatement(// your DML statement);
connection.beginTransaction();
s.execute()
connection.endTransaction();
```

ChengyuanZhao: great point, i added a fix and test.

dzou: Minor point - i would suggest moving to if/else statement now; I feel like the ternary style is no longer easy to read.

elefeint: Seconded. If it looks more like a 3 story condo than like a single-story ranch house, it's time to break that ternary up :D

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
```diff
@@ -136,6 +136,51 @@ public void executeStatementReturnsWorkingStatementWithCorrectQuery() {
...
+        = new SpannerConnection(this.mockClient, TEST_SESSION, TEST_CONFIG);
+    String sql = "insert into books values (title) @title";
+
+    when(this.mockClient.executeBatchDml(any(), any(), any(), any())).thenReturn(Mono.empty());
+
+    StepVerifier.create(
+        Mono.fromSupplier(() -> connection.createStatement(sql))
+            .delayUntil(s -> connection.beginTransaction())
+            .doOnSuccess(SpannerStatement::execute)
+    ).consumeNextWith(x -> {
```
elefeint: I think you can use `expectNextCount(1)` here.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -138,10 +140,9 @@ public Batch createBatch() {
 
   @Override
   public SpannerStatement createStatement(String sql) {
-    SpannerStatement statement
-        = new SpannerStatement(this.client, this, sql, this.config);
-
-    return statement;
+    return getTransactionId() == null && StatementParser.getStatementType(sql) == StatementType.DML
```
dzou: There might be another minor problem here - 

What if someone begins a transaction, creates a statement, commits the transaction, then executes the statement? In such a case, I assume the statement should run in autocommit mode, however they would not have an instance of IndependentSpannerStatement.

```
connection.beginTransaction
Statement s = connection.createStatement()
connection.commit()
s.execute()
```

dzou: It might be better just to have 1 Spanner Statement object, and then you just have a check in the `execute()` method to see if you need to begin/commit a transaction or not.

ChengyuanZhao: This behaves as expected @dzou . the transaction ID would be null and it would run independently.

ChengyuanZhao: fixed!

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
```diff
@@ -136,6 +139,76 @@ public void executeStatementReturnsWorkingStatementWithCorrectQuery() {
...
+    SpannerConnection connection
+        = new SpannerConnection(this.mockClient, TEST_SESSION, TEST_CONFIG);
+    String sql = "insert into books values (title) @title";
+
+    StepVerifier.create(connection.beginTransaction()
+        .doOnSuccess(avoid -> {
+          Statement statement = connection.createStatement(sql);
+          assertThat(statement.getClass()).isEqualTo(AutoCommitSpannerStatement.class);
+        })).verifyComplete();
+    verify(this.mockClient, times(1)).beginTransaction(eq(TEST_SESSION_NAME), any());
```
meltsufin: Why is this relevant? You're not calling execute in this test.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
```diff
@@ -136,6 +139,76 @@ public void executeStatementReturnsWorkingStatementWithCorrectQuery() {
...
     verify(this.mockClient).executeStreamingSql(any(StatementExecutionContext.class), eq(sql),
         eq(EMPTY_STRUCT), eq(EMPTY_TYPE_MAP));
+
+    // Single use READ query doesn't need these round trips below.
+    verify(this.mockClient, times(0)).beginTransaction(eq(TEST_SESSION_NAME), any());
+    verify(this.mockClient, times(0)).commitTransaction(eq(TEST_SESSION_NAME), any());
+  }
+
+  @Test
+  public void executeSingleUseDml() {
```
meltsufin: maybe a more descriptive method name?
something like `executeSingleUseDmlStatementTypeSelection`

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
```diff
@@ -136,6 +139,76 @@ public void executeStatementReturnsWorkingStatementWithCorrectQuery() {
...
+    SpannerConnection connection
+        = new SpannerConnection(this.mockClient, TEST_SESSION, TEST_CONFIG);
+    String sql = "insert into books values (title) @title";
+
+    Statement statement = connection.createStatement(sql);
+    assertThat(statement).isInstanceOf(AutoCommitSpannerStatement.class);
+  }
+
+  @Test
+  public void executeDmlInTransactionTest() {
```
meltsufin: Again, maybe a more descriptive name because you're not really testing `execute`.

meltsufin:
@ChengyuanZhao Just noticed it got merged while I was reviewing. Anyway, if you agree with the comments, maybe a follow-up PR.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
ChengyuanZhao: fixed these in #136 

(things got moved around quite a bit during this PR, these names made sense at one point but then the tests got changed up and I overlooked renaming them)

PR #140, opened 2019-06-20
Add docs about dropped sessions/connections.

### LINKED ISSUE #96: Handle dropped Sessions
Spanner can drop sessions, so Connection should be able to recreate a session on a failed action. 
Figure out if the failed action should be retried on a stale session.

Spanner also recommends [sending keepalive queries](https://cloud.google.com/spanner/docs/sessions#keep_an_idle_session_alive), so perhaps that should be a default + connection setting.

dzou:
Adds docs on dropped sessions.

Fixes #96.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,9 +170,9 @@ it is preferable to reuse your `Connection` object to run multiple statements ra
 a new `Connection` for each statement you wish to run.
 
 Additionally, if a `Connection` is not used for more than 1 hour, the Cloud Spanner database service
-reserves the right to drop the connection. If this occurs, an exception will be thrown when you
-attempt to run queries using the connection, and you will have to recreate the connection in order
-to reattempt the query.
+reserves the right to drop the connection. If this occurs, a `R2dbcNonTransientException` will be
+thrown when you attempt to run queries using the connection, and you will have to recreate the
+connection in order to reattempt the query.
 
 If you definitely need to keep an idle connection alive, for example, if a significant near-term
 increase in database use is expected, then you can prevent the connection from being dropped.
--- a/README.md
+++ b/README.md
@@ -159,6 +159,26 @@ Null values mapping is supported in both directions.
 
 See [Cloud Spanner documentation](https://cloud.google.com/spanner/docs/data-types) to learn more about Spanner types.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -159,6 +159,26 @@ Null values mapping is supported in both directions.
...
+The R2DBC Cloud Spanner `Connection` object represents a persistent connection to a Spanner
+database.
+
+When you instantiate a `Connection` object using the `ConnectionFactory`, a [Spanner session](https://cloud.google.com/spanner/docs/sessions)
+is created and encapsulated within the connection. Creating a session is typically expensive, so
+it is preferable to reuse your `Connection` object to run multiple statements rather than create
+a new `Connection` for each statement you wish to run.
+
+Additionally, if a `Connection` is not used for more than 1 hour, the Cloud Spanner database service
+reserves the right to drop the connection. If this occurs, a `R2dbcNonTransientException` will be
```
meltsufin: I will sound like a broken record, but...
Do code samples make sense here to show how to handle dropped connections?

dzou: Hmm, so I've been thinking over the sample code, it is going to be complex. We actually cannot use the reactor retry operators because we must recreate the entire connection, not simply retry the stream.

I think I would like to revisit the driver handling the dropped session situation again.

README.md
```diff
@@ -159,6 +159,26 @@ Null values mapping is supported in both directions.
...
+thrown when you attempt to run queries using the connection, and you will have to recreate the
+connection in order to reattempt the query.
+
+If you definitely need to keep an idle connection alive, for example, if a significant near-term
+increase in database use is expected, then you can prevent the connection from being dropped.
+Perform an inexpensive operation such as executing the SQL query `SELECT 1` to keep the
```
elefeint: Perhaps we should also implement a "keepalive" setting in the driver itself.

README.md
meltsufin: Maybe we should consult Mark on this. But, yeah, let's revisit in a separate PR.

README.md
dzou: Not a bad idea, thanks.

PR #142, opened 2019-06-20
Allow URLs to be passed through config

### LINKED ISSUE #30: Support URL-based connection
From Mark's [comment](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/26/files#r282182581) on #26:

```
There are two sides to configuration: Programmatic and URL-based. Both ways end up with providing a ConnectionFactoryOptions object.

Maybe you should start drawing of a R2DBC URL for Spanner would look like.

How about r2dbc:spanner:<project>//<instance>/<database>?

This way, <project> gets auto-propagated to ConnectionFactoryOptions.PROTOCOL, <instance> to HOST and <database> to DATABASE.
```

dzou:
Simple propagation for a Cloud Spanner R2DBC URL to be passed through config.

Fixes #30.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,13 +130,29 @@ The following options are available to be configured for the connection factory:
 | Option Name | Description                | Required | Default Value |
 |-------------|----------------------------|----------|---------------|
 | `driver`    | Must be "spanner"          | True     |               |
-| `project`   | Your GCP Project ID        | True     |               |
-| `instance`  | Your Spanner Instance name | True     |               |
-| `database`  | Your Spanner Database name | True     |               |
+| `project`   | Your GCP Project ID        | True (if `url` not provided) | |
+| `instance`  | Your Spanner Instance name | True (if `url` not provided) | |
+| `database`  | Your Spanner Database name | True (if `url` not provided) | |
+| `url`       | A Cloud Spanner R2DBC URL specifying your Spanner database. An alternative to specifying `project`, `instance`, and `database` | False |
 | `google_credentials` | Optional [Google credentials](https://cloud.google.com/docs/authentication/production) override to specify for your Google Cloud account. | False | If not provided, credentials will be [inferred from your runtime environment](https://cloud.google.com/docs/authentication/production#finding_credentials_automatically).
 | `partial_result_set_fetch_size` | Number of intermediate result sets that are buffered in transit for a read query. | False | 1 |
 | `ddl_operation_timeout` | Duration in seconds to wait for a DDL operation to complete before timing out | False | 600 seconds |
 | `ddl_operation_poll_interval` | Duration in seconds to wait between each polling request for the completion of a DDL operation | False | 5 seconds |
+
+### Connection URLs
+
+You may elect to connect to your Spanner database the `url` property instead of specifying the
+`project`, `instance`, and `database` properties separately.
+
+A Cloud Spanner R2DBC URL is constructed using the following format:
+
+```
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -130,14 +130,30 @@ The following options are available to be configured for the connection factory:
 | Option Name | Description                | Required | Default Value |
 |-------------|----------------------------|----------|---------------|
 | `driver`    | Must be "spanner"          | True     |               |
-| `project`   | Your GCP Project ID        | True     |               |
-| `instance`  | Your Spanner Instance name | True     |               |
-| `database`  | Your Spanner Database name | True     |               |
+| `project`   | Your GCP Project ID        | True (if `url` not provided) | |
+| `instance`  | Your Spanner Instance name | True (if `url` not provided) | |
+| `database`  | Your Spanner Database name | True (if `url` not provided) | |
+| `url`       | A Cloud Spanner R2DBC URL specifying your Spanner database. An alternative to specifying `project`, `instance`, and `database` | False |
```
elefeint: "An alternative to specifying `project`, `instance`, and `database` **individually**"?
I don't know, I am missing something at the end of that sentence to show that `url` is instead of **all of those** as opposed to **any of those**. "Individually" does not do it either. Ideas?

README.md
```diff
@@ -130,14 +130,30 @@ The following options are available to be configured for the connection factory:
...
 | `google_credentials` | Optional [Google credentials](https://cloud.google.com/docs/authentication/production) override to specify for your Google Cloud account. | False | If not provided, credentials will be [inferred from your runtime environment](https://cloud.google.com/docs/authentication/production#finding_credentials_automatically).
 | `partial_result_set_fetch_size` | Number of intermediate result sets that are buffered in transit for a read query. | False | 1 |
 | `ddl_operation_timeout` | Duration in seconds to wait for a DDL operation to complete before timing out | False | 600 seconds |
 | `ddl_operation_poll_interval` | Duration in seconds to wait between each polling request for the completion of a DDL operation | False | 5 seconds |
 
+### Connection URLs
+
+You may elect to connect to your Spanner database the `url` property instead of specifying the
```
elefeint: insert "using" before "the `url` property"

dzou:
> If the URL is malformed (project missing, for example), what does failure look like? Spanner probably throws some form of 404. Can we do better and analyze when URL is in a bad format?

Sure, that is reasonable. The error messages are indeed cryptic.

If you omit the project ID, you will get this message: `INVALID_ARGUMENT: Invalid CreateSession request.`

README.md
meltsufin:
```suggestion
You may specify the coordinates of your Cloud Spanner database using the `url` property instead of specifying the
```

Btw, should we enforce Artem's one-sentence-per-line rule? 😉

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfigurationTest.java
```diff
@@ -129,4 +129,16 @@ public void ddlOperationWaitSettings() {
     assertThat(config.getDdlOperationTimeout()).isEqualTo(Duration.ofSeconds(23));
     assertThat(config.getDdlOperationPollInterval()).isEqualTo(Duration.ofSeconds(45));
   }
+
+  @Test
+  public void databaseUrlSettings() {
```
meltsufin: Can we have a test that compares URL vs. project-database-instance properties to see if they are equivalent?

README.md
dzou: Hmm ended up going with "separately" though I echo your concerns. Hopefully this will be clear enough.

Issue #144, opened 2019-06-21
Refactor GrpcClient to be per connection instead of singleton


meltsufin:
Currently we carry around execution context along with the `Client` to the statement to allow it to execute itself. We can instead just add the execution context to the `Client` itself. This will require creating a new `GrpcClient` per connection.

Also, currently the statement may start a transaction for auto-commit. Statements should not be responsible for starting transactions. For auto-commit, the `Client` should automatically start and commit the transaction without making the `Statement` aware of it.

Follow-up to discussion in https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/131#discussion_r295881725

meltsufin:
Closing following discussion with Dan.

PR #146, opened 2019-06-24
Refactor Statement and Batch to not manage Transaction

dzou:
Refactors the Statement and Batch classes to not manage begin/commit of transaction. This is delegated to grpcClient.

Contributes to #144.


src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
```diff
@@ -141,71 +138,23 @@ public void executeStatementReturnsWorkingStatementWithCorrectQuery() {
...
-            .doOnSuccess(SpannerStatement::execute)
-    ).consumeNextWith(x -> {
-    }).verifyComplete();
+            .doOnSuccess(SpannerStatement::execute))
+        .consumeNextWith(x -> {})
+        .verifyComplete();
     verify(this.mockClient, times(1)).beginTransaction(eq(TEST_SESSION_NAME), any());
   }
 
-  @Test
```
ChengyuanZhao: I think you should keep this one. It does a manual transaction followed by a separate execution even though the statement was created inside the transaction.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -191,56 +198,40 @@ public GrpcClient(GoogleCredentials credentials) {
...
-      }
+      ExecuteBatchDmlRequest.Builder request =
+          ExecuteBatchDmlRequest.newBuilder()
+              .setSession(ctx.getSessionName())
+              .addAllStatements(statements)
+              .setSeqno(ctx.nextSeqNum());
 
-      return ObservableReactiveUtil
-          .unaryCall(obs -> this.spanner.executeBatchDml(request.build(), obs));
+      if (ctx.getTransactionId() != null) {
```
ChengyuanZhao: Hmmm there's a difference here it seems.

So the `ctx.getTransactionId() != null` here happens inside the mono.defer of the response, which means this context isn't checked until after the user tries to consume the result mono. 

Previously we were checking this condition on creating the mono (not retrieving the results from the mono). 

I'm not quite sure which one is correct, but that's definitely a difference.

dzou: Are you sure that is correct? I think it may be the same. In the previous code, the `ctx.getTransactionId() != null` check was done in the `createBatchDmlRequestBuilder()` method; that lives inside the Mono.defer.

ChengyuanZhao: I was looking at the one in `AutoCommitStatement`. That one is done when statement.execute() is called, which is outside of the mono.defer

src/main/java/com/google/cloud/spanner/r2dbc/StatementExecutionContext.java
```diff
@@ -34,7 +34,7 @@
    * Retrieves the session name associated with the current connection.
    * @return valid name of a currently valid Cloud Spanner session.
    */
-  public String getSessionName();
```
elefeint: I was thinking before that acquiring a `StatementExecutionContext` and figuring out the current session name and other state might be useful for user applications.
Not a strong opinion though.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -191,56 +198,40 @@ public GrpcClient(GoogleCredentials credentials) {
...
+        request.setTransaction(TransactionSelector.newBuilder().setId(ctx.getTransactionId()));
+        return ObservableReactiveUtil.unaryCall(
+            obs -> this.spanner.executeBatchDml(request.build(), obs));
+      } else {
+        return beginTransaction(ctx.getSessionName(), READ_WRITE_TRANSACTION)
```
elefeint: Behold, I am about to argue with myself:
* point: we don't have to do a separate `BeginTransaction` call because[ `ExecuteBatchDmlRequest` ](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.ExecuteBatchDmlRequest)can contain a transaction selector asking to start a new transaction. We'd still need to commit separately.
* counter-point: we'd then have to extract the ID of the newly created transaction from an unusual place: " The transaction ID of the new transaction is returned in ResultSetMetadata.transaction, which is a Transaction.".

src/main/java/com/google/cloud/spanner/r2dbc/StatementExecutionContext.java
dzou: O I didn't delete this method; I just removed the `public` modifier from it.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
dzou: I couldn't keep this test in its current form because the begin/commit logic is now wrapped in the client, so we can't count the mocked calls on the client (since it is now an internal function of the client calling begin/commit on itself).

I added a new test to SpannerIT to cover case of creating a statement in a transaction and executing outside.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
dzou: I see. In that case I think the current way is preferable because the statement no longer needs to know about the transaction state.

In the previous way, we could probably construct a case where a user builds a full stream (connection.createStatement().execute()) when a transaction is active but then subscribe to it after the transaction is committed; this might be another case which breaks the previous impl.

ChengyuanZhao: Actually I think this is an illustration why it is worse.

I think the user's intention is defined by how he writes the code that creates the mono. In other words, if the user calls "execute" on the statement between "beginTransaction" and "commitTransaction" that means he wants this statement to execute inside the transaction (not as auto-commit). This .execute call is BEFORE the mono is consumed, so we need to know the ctx state at the mono creation time. 

If the user then consumes this execute result later then according to your current code it would execute as a stand-alone , which is NOT what he intends in my opinion. He should be told "you tried to do this DML inside a transaction, but in fact that transaction has ended and you tried consuming it too late."

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
dzou: Hmm I see. I think this is a good point; it's probably preferable to do it this way to avoid an extra round-trip; I think I'll do it.

src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionTest.java
ChengyuanZhao: i see

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
dzou: Hmm, I see your point. Perhaps this discussion is ultimately moot because the user will be using the pattern of chaining their `begin/execute/commit` Monos together; this edge-case we are discussing is contrived.

Overall I think it is the reactive paradigm to do all your work inside a Mono.defer and not immediately, at least this seemed to be the heuristic/style used, so I am still inclined to keep it this way.

ChengyuanZhao: But isn't it actually kinda easy to do this?

```

factory.create()
.delayUntil(c->c.beginTransaction)
.flatMapMany( c-> c.createStatement("SELECT accounts to transfer here..." ).execute())
.map(c-> c.createSTatement( [delete money from account here]).execute())
.delayUntil(c-> c.commitTransaction())
.map(c-> print("let's see how much money we just deleted: " + c.getRowsUpdated());

// right there you would be deleting money from accounts non-atomically in a separate condition just because you don't check the results until after the commit 
```

And while it's easy to make a blanket statement to say they prefer doing everything in a mono.defer, there's the implicit "if it doesn't change the actual logic" at the end!

dzou: Hmm, ok, in that case it should return an error right? Instead of run? Yeah I see your point.
I would want to ask mark to understand what he intended for this interface just to confirm. The mental block for me still is that I feel that the statement should not need to know about transaction state.

Also note this example actually is protected against in our driver. I think due to idiosyncrasy of Spanner we had to make it so DML queries will actually execute on `execute()`; however select queries will defer the results are retrieved.

dzou: Hey @mp911de -

We have a question about the behavior of `Statement.execute()`

In the below example, the user starts a transaction to execute a *read* and then *write* on a table. However, the user by mistake calls `c.getRowsUpdated()` outside the `commitTransaction()` on the second statement. As we understand it, the user must get the results of a query for it to be fully run in Reactive; thus, the user would execute the statement outside the transaction in the below example.

```
factory.create()
.delayUntil(c->c.beginTransaction)
.flatMapMany( c-> c.createStatement("SELECT accounts to transfer here..." ).execute())
.map(c-> c.createSTatement( [delete money from account here]).execute())
.delayUntil(c-> c.commitTransaction())
.map(c-> print("let's see how much money we just deleted: " + c.getRowsUpdated());

// right there you would be deleting money from accounts non-atomically in a separate condition just because you don't check the results until after the commit
```

In this case, when the statement is run: Should the statement be run as a stand-alone, autocommit statement? Or should an error be thrown that the transaction no longer exists?

mp911de: @dzou that behavior is not defined. What typically happens is that  `.execute()` participates in the transaction and you get a `Result` object that is attached to the transaction. Depending on whether you have all results in memory, or whether you need to fetch data, behavior can be different. If you have all results in memory and you call `getRowsUpdated()` after committing the transaction, then you could return the cached data.

If you need to issue another query/talk to the server, after the transaction is committed, then probably you would run into an error.

It's fair to say that any object associated with a transaction gets invalidated if the transaction gets committed. This is usual for SQL databases (e.g. close a result set when the transaction gets cleaned up or invalidate BLOB/CLOB objects after a transaction).

Does this help?

dzou: Thanks Mark, that is helpful.

@ChengyuanZhao - So to resolve this discussion, would it be reasonable to just move forward with this PR and then create an issue to revisit this edge case later? The reason for deferring this would be that 1. our driver behaves correctly because we run DML queries on `execute()` 2. I would need to think about what are the correct changes to introduce to make the SpannerStatement track the transaction state.

ChengyuanZhao: sounds good to me!

dzou: Ok, created: #150.

PR #147, opened 2019-06-24
Fix Javadoc Lint Errors

### LINKED ISSUE #135: Fix javadoc errors
Currently javadoc generation is turned on under the `release` profile. However, linting is turned off using ` <additionalparam>-Xdoclint:none</additionalparam>` because we have errors in our javadocs. 

We should fix javadoc issues and re-enable javadoc linting. Also, we should enable javadoc generation on Travis to verify that errors are not introduced in the future, after this is fixed.

dzou:
Fix the Javadoc lint errors.

Fixes #135.


.travis.yml
```diff
@@ -23,3 +23,4 @@ script:
     else
       ./mvnw verify -B -V -DskipITs;
     fi;
+  - ./mvnw clean javadoc:jar
```
meltsufin: If you do something like this, you'll want to make sure you trigger the specific javadoc configuration.
```suggestion
  - ./mvnw clean javadoc:jar@attach-javadocs
```

dzou: Done.

dzou:
I think this should work now; here's the build with a failed javadoc change: https://travis-ci.com/GoogleCloudPlatform/cloud-spanner-r2dbc/builds/116863304

Now we'll wait to see if the build passes after fixing the javadoc error.

dzou:
Had to resolve a merge conflict; PTAL.

PR #149, opened 2019-06-26
change timestamp type to LocalDateTime

### LINKED ISSUE #100: Data Type mapping compliance with R2DBC spec
We need to ensure compliance with the data type mapping as [defined in the spec](https://r2dbc.io/spec/0.8.0.M8/spec/html/#datatypes), and recently amended in https://github.com/r2dbc/r2dbc-spi/pull/100.

If we are unable to comply with any of it, it should be documented.

dmitry-s:
fixes #100

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,9 +171,6 @@ Cloud Spanner R2DBC Driver supports the following types:
 |`TIMESTAMP`     |`java.time.ZonedDateTime` |
 |`ARRAY`         |Array-Variant of the corresponding Java type (e.g. `Long[]` for `ARRAY<INT64>`)|
 
-Note that Cloud Spanner Timestamp type stores time in UTC zone.
-Because of that timestamp value is converted to UTC on write, so when it is read, timestamp values will always be in UTC. 
-
 Null values mapping is supported in both directions.
 
 See [Cloud Spanner documentation](https://cloud.google.com/spanner/docs/data-types) to learn more about Spanner types.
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/cloud/spanner/r2dbc/codecs/DefaultCodecsTest.java
```diff
@@ -97,13 +115,16 @@
   @Parameter(2)
   public Type valueType;
 
+  @Parameter(3)
+  public Object decodedVal;
```
meltsufin: Can you avoid this?
There should be a way to check if two timestamps are equivalent regardless of their timezone.

dmitry-s: good idea

ChengyuanZhao:
@meltsufin @dmitry-s the only difference between the two is that `ZonedDateTime` knows about things like daylight savings time. A "zone" is an "offset" plus daylight savings time adjustment. 

It makes no difference to us because Spanner can store neither an offset nor a time zone. In either case we can only create an object using either 0 offset or "UTC" zone. The UTC time zone has no DST adjustments, so it is the same as 0 offset always.

README.md
```diff
@@ -168,7 +168,7 @@ Cloud Spanner R2DBC Driver supports the following types:
 |`INT64`         |`java.lang.Long`     |
 |`INT64`         |`java.lang.Integer`  |
 |`STRING`        |`java.lang.String`   |
-|`TIMESTAMP`     |`java.sql.Timestamp` |
+|`TIMESTAMP`     |`java.time.ZonedDateTime` |
```
meltsufin: Still says zoned.

PR #151, opened 2019-06-26
Correctly propagate errors from Batch DML

### LINKED ISSUE #148: DML queries must check the response object for error codes.
The `ExecuteBatchDml` method returns a response object that has a `Status` field describing success/failure of the operation. Failure need to be parsed and propagated to the end user.

https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#executebatchdmlresponse

dzou:
Make DML queries propagate error from the response object if present.

[ExecuteBatchDmlResponse contains a Status](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#executebatchdmlresponse) which contains message if DML query fails. 

Fixes #148

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
@@ -52,6 +52,7 @@ import com.google.spanner.v1.Type;
 import io.grpc.CallCredentials;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
+import io.grpc.Status;
 import io.grpc.auth.MoreCallCredentials;
 import io.r2dbc.spi.R2dbcNonTransientResourceException;
 import java.time.Duration;
@@ -229,6 +230,14 @@ public class GrpcClient implements Client {
                 return Mono.empty();
               }
             });
+      }
+    })
+    .handle((response, sink) -> {
+      if (response.hasStatus() && response.getStatus().getCode() != Status.Code.OK.value()) {
+        sink.error(
+            new R2dbcNonTransientResourceException(response.getStatus().getMessage()));
+      } else {
+        sink.next(response);
       }
     });
   }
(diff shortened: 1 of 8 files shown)


dzou:
Not quite ready just yet; I realized this needs to be tweaked to correctly handle the case where some statements succeed and the last one fails.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -230,6 +231,14 @@ public GrpcClient(GoogleCredentials credentials) {
               }
             });
       }
+    })
+    .handle((response, sink) -> {
+      if (response.hasStatus() && response.getStatus().getCode() != Status.Code.OK.value()) {
+        sink.error(
+            new R2dbcNonTransientResourceException(response.getStatus().getMessage()));
```
dmitry-s: can we return status code to the user as well?

dzou: We could but that would require us creating new exception objects for which I think we ought defer to a later PR. Created: #154

src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerIT.java
```diff
@@ -260,6 +260,14 @@ public void testRunDmlAfterTransaction() {
         .verifyComplete();
   }
 
+  @Test
+  public void testDmlExceptions() {
+    StepVerifier.create(
+        Mono.from(connectionFactory.create())
+            .flatMapMany(conn -> conn.createStatement("INSERT BOOKS asdfasdfasdf").execute()))
```
ChengyuanZhao: love that book.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -200,7 +202,7 @@ public GrpcClient(GoogleCredentials credentials) {
   }
 
   @Override
-  public Mono<ExecuteBatchDmlResponse> executeBatchDml(
+  public Flux<ResultSet> executeBatchDml(
```
dmitry-s: Why is it Flux now?

dzou: Had to make it flux in order to append the error to the end of the resultSets flux if the status != OK. This is equivalent to the previous pattern of doing the flatMapIterable immediately after calling `executeBatchDml`.

dzou: Also, just added the test we discussed: https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/151/commits/ca7e157d6c9dcf0708a88900e5632b9ee2c8b085

PTAL.

PR #153, opened 2019-06-27
Clarify backpressure concepts in the docs.

### LINKED ISSUE #152: Clarify prefetch and backpressure concepts
Based on my research, the concept of prefetch and backpressure for cloud spanner is a little more nuanced than we initially wrote in the documentation.

See: https://stackoverflow.com/questions/56675511/what-does-prefetch-mean-in-project-reactor

> Prefetch will affect how many items (PartialResultSets) Reactor requests from the Publisher in it's first request. It is an upper bound that affects subsequent requests where subsequent requests will be triggered when 75% of the prefetch amount has been emitted so it will not increase automatically.

dzou:
Clarify concepts regarding backpressure and prefetch in the documentation.

Based on my research, the fetch-size setting is a bit more nuanced and could be additionally clarified.
Updated the docs with more information.

See:
https://stackoverflow.com/questions/56675511/what-does-prefetch-mean-in-project-reactor
https://projectreactor.io/docs/core/release/reference/#_on_backpressure_and_ways_to_reshape_requests

Fixes #152.

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,11 +332,23 @@ See the above section regarding `ConnectionFactory` options for more information
 
 ## Back Pressure
 
-Table rows are transmitted from Cloud Spanner in fragments called `PartialResultset`.
-The number of fragments for each row cannot be determined beforehand. 
-While you can decide the number of rows you request from `SpannerResult`, the Cloud Spanner R2DBC driver will always request a fixed number of fragments from Cloud Spanner to fulfill your request and will do so repeatedly if necessary.
+When you execute a **read** query in Cloud Spanner, the table rows of the result are transmitted
+back to the client in chunks called `PartialResultSets`. Every `PartialResultSet` object may
+contain any number of complete (or incomplete) table rows, and the number of rows cannot be
+determined beforehand. Consequently, the driver requests these `PartialResultSet` objects from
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -332,11 +332,23 @@ See the above section regarding `ConnectionFactory` options for more information
...
+contain any number of complete (or incomplete) table rows, and the number of rows cannot be
+determined beforehand. Consequently, the driver requests these `PartialResultSet` objects from
+upstream and holds a buffer of these objects from which the table rows of the results are extracted.
+As the buffer is depleted, more `PartialResultSets` will be streamed from Cloud Spanner and
+will replenish the buffer.
+
+In order to support backpressure under these conditions, the driver provides a setting
+`partial_result_set_fetch_size` which specifies how many PartialResultSet objects the driver
+requests from Spanner in it's first request to store in the buffer. This number is an upper
+bound that affects how it is replenished; when the client has consumed 25% of the PartialResultSets,
```
ChengyuanZhao: I think this sentence about the 25% is necessary. The previous half of this paragraph and the preceding paragraph say the exact same thing as before but is just more wordy.

ChengyuanZhao: Eh actually, I'm fine with it. it's different but not actually worse. LGTM

PR #156, opened 2019-07-02
Keep alive

### LINKED ISSUE #141: Implement keepalive option
Spanner recommends to run a simple query (`SELECT 1`) periodically to avoid sessions being dropped.

The driver could do that as an optional feature, configured through connection properties.

dmitry-s:
fixes #141

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
@@ -25,11 +25,15 @@ import com.google.spanner.v1.TransactionOptions.ReadWrite;
 import io.r2dbc.spi.Batch;
 import io.r2dbc.spi.Connection;
 import io.r2dbc.spi.IsolationLevel;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import javax.annotation.Nullable;
 import org.reactivestreams.Publisher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 /**
@@ -41,6 +45,8 @@ public class SpannerConnection implements Connection, StatementExecutionContext 
       TransactionOptions.newBuilder()
           .setReadWrite(ReadWrite.getDefaultInstance())
           .build();
+
+  private static final String KEEP_ALIVE_STATEMENT = "SELECT 1";
 
   private final Logger logger = LoggerFactory.getLogger(this.getClass());
 
@@ -56,6 +62,8 @@ public class SpannerConnection implements Connection, StatementExecutionContext 
 
   private final SpannerConnectionConfiguration config;
 
+  private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+
   /**
    * Instantiates a Spanner session with given configuration.
    * @param client client controlling low-level Spanner operations.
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -56,6 +62,8 @@
 
   private final SpannerConnectionConfiguration config;
 
+  private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
```
meltsufin: Why create the pool if the setting is potentially not enabled?
Also, we probably shouldn't create a pool per connection.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -66,6 +74,21 @@ public SpannerConnection(Client client, Session session, SpannerConnectionConfig
     this.client = client;
     this.session = session;
     this.config = config;
+
+    if (config != null && config.getKeepAliveEnabled()) {
+      this.scheduler.scheduleAtFixedRate(
+          () -> {
+            if (this.transaction != null && this.transaction.isInitialized()) {
```
meltsufin: Is this thread safe?
Also, the keep alive only works if there is no active transaction?

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -156,6 +172,10 @@ public int hashCode() {
 
     private Duration ddlOperationPollInterval = Duration.ofSeconds(5);
 
+    private Boolean keepAliveEnabled = true;
+
+    private Duration keepAliveInterval = Duration.ofMinutes(10);
```
dzou: I don't think the interval should be configurable because a session will only be deleted if it has been idle for 1 hour; thus the interval should just be as large as possible: 1 hour (or 59 mins. to be safe). A user would want to minimize the number of keep-alive queries they make.

Implementation could be improved by resetting the timer each time a user executes a query to further minimize keep-alive queries sent.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -66,6 +74,21 @@ public SpannerConnection(Client client, Session session, SpannerConnectionConfig
...
+              return;
+            }
+            Flux.from(Flux.from(createStatement(KEEP_ALIVE_STATEMENT).execute())
+                .collectList()
+                .block().get(0)
+                .map((row, rowMetadata) -> 0)).subscribe();
```
dzou: I think it would be better to invent a `SessionProvider` object in which you encapsulate the logic of keep-alive. And every time you called `sessionProvider.get()`, if you find that your session was dropped for some reason, it can create a new session for the connection behind the scenes.

dmitry-s:
This functionality exists in [r2dbc-pool](https://github.com/r2dbc/r2dbc-pool)

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
dmitry-s: Where did you get the 1h interval? What I see in the docs is 

> The Cloud Spanner database service reserves the right to drop an unused session. 

Keep-alive statements are extremely cheap, there is really no point in minimizing the number of them when if we don't know when the session is going to be dropped.

ChengyuanZhao: Yea I've had connections end themselves because I was a few seconds too slow while stepping through code while testing..

PR #160, opened 2019-07-15
Propagate Specific R2dbc Exceptions by Error Code

### LINKED ISSUE #154: Propagate Error Status Code through Exceptions
Currently, we are using `R2dbcTransientException` and `R2dbcNontransientException` to propagate exceptions to the user.

These exceptions do not contain error codes. So we would like to extend a new exception type which specifies transience but also propagate error codes to the user.

context: https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/151#discussion_r298242609

dzou:
If a Spanner-specific exception is thrown, this adds support to extract the error code if present and construct the appropriate R2dbc exception subclass based upon the error code.

Fixes #154.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
@@ -243,8 +243,7 @@ public class GrpcClient implements Client {
         R2dbcException exception =
             SpannerExceptionUtil.createR2dbcException(
                 response.getStatus().getCode(),
-                response.getStatus().getMessage(),
-                null);
+                response.getStatus().getMessage());
         results = results.concatWith(Mono.error(exception));
       }
       return results;
--- a/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
@@ -20,6 +20,7 @@ import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.spanner.r2dbc.StatementExecutionContext;
 import com.google.cloud.spanner.r2dbc.util.Assert;
 import com.google.cloud.spanner.r2dbc.util.ObservableReactiveUtil;
+import com.google.cloud.spanner.r2dbc.util.SpannerExceptionUtil;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.longrunning.GetOperationRequest;
 import com.google.longrunning.Operation;
@@ -55,6 +56,7 @@ import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.Status;
 import io.grpc.auth.MoreCallCredentials;
+import io.r2dbc.spi.R2dbcException;
 import io.r2dbc.spi.R2dbcNonTransientResourceException;
 import java.time.Duration;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -238,9 +240,11 @@ public GrpcClient(GoogleCredentials credentials) {
     .flatMapMany(response -> {
       Flux<ResultSet> results = Flux.fromIterable(response.getResultSetsList());
       if (response.hasStatus() && response.getStatus().getCode() != Status.Code.OK.value()) {
-        results = results.concatWith(
-            Mono.error(
-                new R2dbcNonTransientResourceException(response.getStatus().getMessage())));
+        R2dbcException exception =
```
meltsufin: You don't really need this local variable, but if you added it for readability, that's fine.

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
```diff
@@ -32,37 +39,81 @@
...
           "Received unexpected EOS on DATA frame from server");
 
+  /**
+   * Manually creates an {@link R2dbcException} from a provided error code and message.
+   *
+   * @param errorCode the Spanner error code of the error.
+   * @param message the error message.
+   * @return the resulting {@link R2dbcException} that is propagated to the user.
+   */
+  public static R2dbcException createR2dbcException(int errorCode, String message) {
```
meltsufin: Seeing how you added a bunch of these `create` methods, would it make sense to just put this logic into `R2dbcExcpetion` constructors instead?
In fact, refactoring this whole class into `R2dbcException` could be an option.

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
dzou: Yup, I just had it for readability.

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
dzou: Ah, I don't think I can quite do that since the R2dbcException classes are owned by the SPI; not controlled by us.

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
```diff
@@ -32,37 +39,81 @@
...
+  }
+
+  /**
+   * Extracts metadata of a thrown exception and creates a {@link R2dbcException} from it.
+   *
+   * @param baseException the base exception that is thrown
+   * @return the resulting {@link R2dbcException} that is propagated to the user.
+   */
+  public static R2dbcException createR2dbcException(Throwable baseException) {
+    if (!(baseException instanceof StatusRuntimeException)) {
```
elefeint: I'd check for null `baseException`, just in case.

dzou: Done.

src/main/java/com/google/cloud/spanner/r2dbc/util/SpannerExceptionUtil.java
meltsufin: Subclass it?
`SpannerR2dbcException`?

dzou: But I still have to return the correct subclass of R2dbcException such as: R2dbcDataIntegrityViolationException, R2dbcBadGrammarException, etc. These extend R2dbcException.

meltsufin: Ah, good point. That would be annoying to subclass all of them.

PR #170, opened 2019-11-04
Spanner connection-level metadata

elefeint:
The metadata is trivial:
* database product name is the same as  in `SpannerConnectionFactoryMetadata`. 
* database version is "n/a" as that is not a relevant concept for Cloud Spanner.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionMetadata.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionMetadata.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionMetadata.java
```diff
@@ -0,0 +1,38 @@
...
+public class SpannerConnectionMetadata implements ConnectionMetadata {
+
+  @Override
+  public String getDatabaseProductName() {
+    return SpannerConnectionFactoryMetadata.NAME;
+  }
+
+  @Override
+  public String getDatabaseVersion() {
+    return "n/a";
```
meltsufin: There is a version of the gRPC API, I believe. Maybe that would be relevant. Maybe check with the Cloud Spanner team.

PR #171, opened 2019-11-11
Add ability to turn autocommit on and off

elefeint:
Keeps separate state for autocommit; does not rely on transaction.
A couple of edge case scenarios:
* autocommit was turned off but transaction not started yet. In this case, attempting DML statements will still behave as if autocommit is turned on (a side effect of current GrpcClient.executeBatchDml implementation). I can see an argument for changing this to conform to SPI behavior of committing in-progress transaction upon autocommit being turned off.
* autocommit was turned back on while a non-readwrite transaction is in progress -- this will NOT result in read/write transaction being committed. This is different from the more typical scenario, in which autocommit is turned back on while a Read/Write transaction is in progress, and transaction gets committed automatically. The reason for this is that a read-only transaction guarantees consistent reads, and closing the transaction without user instruction could lead to inconsistent reads. This will never happen through a framework such as Spring Data, as read-only transactions are not part of the R2DBC SPI. Therefore, a user has to intentionally start a read-only transaction by downcasting `Connection` as `SpannerConnection` and using the custom overloaded `beginTransaction(TransactionOptions)` method.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
@@ -227,16 +227,14 @@ public class SpannerConnection implements Connection, StatementExecutionContext 
    */
   @Override
   public Publisher<Void> setAutoCommit(boolean newAutoCommit) {
-    return Mono.defer(() -> {
-      if (newAutoCommit && !this.autoCommit && this.transaction != null) {
-        return this.commitTransaction(false)
-            .doOnSuccess(none -> {
-              this.autoCommit = newAutoCommit;
-            });
-      }
-      this.autoCommit = newAutoCommit;
-      return Mono.empty();
-    });
+    return Mono.defer(
+        () -> {
+          boolean commitNeeded = newAutoCommit && !this.autoCommit && this.transaction != null;
+          return (commitNeeded ? this.commitTransaction(false) : Mono.<Void>empty())
+              .doOnSuccess(none -> {
+                this.autoCommit = newAutoCommit;
+              });
+        });
   }
 
   /**
@@ -245,7 +243,7 @@ public class SpannerConnection implements Connection, StatementExecutionContext 
    *
    * <p>When autocommit is on, each standalone DML query will be executed in its own Read/Write
    * transaction.
-   * 
+   *
    * <p>For batching multiple DML queries, see {@link #createBatch()}.
    *
    * @return whether autocommit mode is on.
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -211,16 +216,43 @@ private void setTransaction(
   }
 
 
+  /**
+   * Changes the autocommit mode of the current connection. No-op if the value is unchanged.
+   *
+   * <p>If autocommit was previously off and a read/write transaction is in progress, the
```
dzou: Sorry I realized I missed making a comment about changing autocommit during a transaction in your doc.

One thing I saw was that in the Spanner JDBC driver by Knut, he will throw an exception if the user tries to change the autocommit during a transaction. What are your thoughts about the pros/cons of this approach versus committing if it was previously off.

elefeint: This one is easy; the[ specification demands it](https://r2dbc.io/spec/0.8.0.RC2/spec/html/#transactions.boundaries): "If the value of auto-commit is changed during an active transaction, the current transaction is committed".

We'll just have to document that this is different than JDBC driver.

src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnection.java
```diff
@@ -211,16 +216,43 @@ private void setTransaction(
...
+   * @param newAutoCommit whether autocommit should be on or off in the future.
+   * @return {@link Mono} of the transaction commit operation, if applicable; empty mono otherwise.
+   */
   @Override
-  public Publisher<Void> setAutoCommit(boolean b) {
-    // TODO: https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/issues/165
-    throw new RuntimeException("Turning off autocommit is not supported yet.");
+  public Publisher<Void> setAutoCommit(boolean newAutoCommit) {
+    return Mono.defer(() -> {
+      if (newAutoCommit && !this.autoCommit && this.transaction != null) {
```
dzou: If you wanted to avoid repeating the `this.autocommit = newAutoCommit` line, would this work?

```
return Mono.defer(() -> {
      Mono<Void> resultMono;

      if (newAutoCommit && !this.autoCommit && this.transaction != null) {
        resultMono = this.commitTransaction(false);
      } else {
        resultMono = Mono.empty();
      }
      return resultMono.doOnSuccess(done -> this.autoCommit = newAutoCommit);
    });
```

elefeint: It works -- I changed it up a bit; take a look.

PR #173, opened 2019-11-15
Implement connection validation

### LINKED ISSUE #162: Implement Connection.validate
R2DBC SPI added a connection validation method on the interface in r2dbc/r2dbc-spi#54

elefeint:
### Behavior change:
* Null out session when connection is closed.
* Local validation relies on having a valid session (it checks for session name, but that's only going to be non-null if and only if the connection is non-null.)
* Remote validation is implemented as a low-level `GrpcClient` healthcheck. It sends a trivial query to Spanner over `executeSql` (as opposed to `executeStreamingSql`).

### TCK test change:
* Removed session pools from TCK tests
* Suppressed error logging when already-existing tables try to get created. This operation also takes forever; need to check whether pre-querying tables will be faster.

Fixes #162 .

### DIFF
--- a/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
+++ b/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
@@ -40,6 +40,7 @@ import com.google.spanner.v1.ExecuteBatchDmlRequest;
 import com.google.spanner.v1.ExecuteBatchDmlRequest.Statement;
 import com.google.spanner.v1.ExecuteBatchDmlResponse;
 import com.google.spanner.v1.ExecuteSqlRequest;
+import com.google.spanner.v1.ExecuteSqlRequest.Builder;
 import com.google.spanner.v1.PartialResultSet;
 import com.google.spanner.v1.ResultSet;
 import com.google.spanner.v1.RollbackRequest;
@@ -61,6 +62,8 @@ import io.r2dbc.spi.R2dbcNonTransientResourceException;
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -81,13 +84,18 @@
 
   private static final String USER_AGENT_LIBRARY_NAME = "cloud-spanner-r2dbc";
 
+  private static final String HEALTHCHECK_SQL = "SELECT 1";
+
   private static final int PORT = 443;
 
+  private final Logger logger = LoggerFactory.getLogger(this.getClass());
```
meltsufin: `static`

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -331,6 +336,30 @@ public GrpcClient(GoogleCredentials credentials) {
...
+      if (ctx.getSessionName() == null) {
+        return Mono.just(false);
+      }
+
+      return ObservableReactiveUtil.<ResultSet>unaryCall(
+          obs -> this.spanner.executeSql(buildSqlRequest(ctx, HEALTHCHECK_SQL).build(), obs)
+      )
+          .map(rs -> Boolean.TRUE)
+          .onErrorResume(error -> {
+            this.logger.warn("Spanner healthcheck failed", error);
```
meltsufin:
```suggestion
            this.logger.warn("Cloud Spanner healthcheck failed", error);
```

src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java
```diff
@@ -331,6 +336,30 @@ public GrpcClient(GoogleCredentials credentials) {
...
+            return Mono.just(Boolean.FALSE);
+          });
+    });
+  }
+
+  private Builder buildSqlRequest(StatementExecutionContext ctx, String sql) {
```
meltsufin: I'm not sure there is a significant benefit to having this method. It's only used in one place right now.

elefeint: Two -- executeStreamingSql() and healthcheck()

elefeint: I am open to inlining it back; the reason I extracted the small bit of functionality is that with reactive style programming, even the smallest bit of boilerplate looks distracting from the flow of operations, what with callbacks and deferreds.

meltsufin: sorry, missed the second usage.

PR #175, opened 2019-11-21
Fix TCK test failures

elefeint:
Two reasons for failures:
1) Cloud Spanner requires that columns be enumerated in insert statement.
2) Cloud Spanner's "Integer" type is 64 bit, so it maps to a Java Long.

Also: renamed from Example to TestKit to conform to new naming convension.

In the long run, it would be nice to fix without overriding each test, hence r2dbc/r2dbc-spi#138.


elefeint:
Watch the SPI repo, I suppose. But that's why I filed that issue -- it's a risk of falling behind for us to maintain a copy .

PR #177, opened 2019-11-22
Make table creation conditional on missing tables

### LINKED ISSUE #174: TCK integration test table creation takes a long time to fail.
Investigate whether it's faster to check for table existence before trying to create them.
Rename TCK tests away from Example while there.

elefeint:
Speeds up TCK test setup.

Fixes #174 .

### DIFF
--- a/src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerTestKit.java
+++ b/src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerTestKit.java
@@ -47,6 +47,7 @@ import java.util.function.Function;
 import java.util.stream.IntStream;
 import org.junit.Ignore;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerTestKit.java
```diff
@@ -103,25 +108,37 @@
...
-    } catch (Exception e) {
-      if (!e.getMessage().contains("Duplicate name in schema")) {
+
+  private static void createTableIfNeeded(DatabaseId id, String tableName, String definition) {
+    Boolean tableExists = Mono.from(connectionFactory.create())
+        .flatMapMany(c -> c.createStatement(
+            "SELECT table_name FROM information_schema.tables WHERE table_name = @name")
+                .bind("name", tableName)
+                .execute())
+        .flatMap(result -> result.map((row, meta) -> true))
```
dzou: Since you must check if any rows are returned to determine if the table is present, is it possible to do a `.hasElements()` to check the result? I.e:

```
Boolean tableExists = Mono.from(connectionFactory.create())
        .flatMapMany(c -> c.createStatement(
            "SELECT table_name FROM information_schema.tables WHERE table_name = @name")
                .bind("name", tableName)
                .execute())
         // .flatMap(result -> result.map((row, meta) -> true))   // not sure if this line is necessary
        .hasElements()
        .block();
```

elefeint: The mapping line was needed because there is always a result set, but there aren't always rows. The flatMap converts from result set to a stream of rows, which could be empty.

Good idea on hasElements!

PR #178, opened 2019-11-22
remove runnable layer

### LINKED ISSUE #161: Simplify delayUntil tests
We have a pattern of `.delayUntil(c -> Mono.fromRunnable(() -> StepVerifier.create(...)))` to work around reactor/reactor-core#1741, which is now resolved.

Once it's released (in 3.2.11 - Californium-SR10?), remove the extra mono wrapper.

elefeint:
I was going to just replace `Mono.fromRunnable()` with `Mono.just(/*step verifier here*/)` but ended up inverting the logic, so that `StepVerifier` is always the outer layer. The test purpose feels easier to understand to me this way, but tell me if you'd rather keep the original way.

Fixes #161.

### DIFF
--- a/src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerIT.java
+++ b/src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerIT.java
@@ -345,8 +345,9 @@ public class SpannerIT {
     ).expectNext(1, 1, 1)
         .verifyComplete();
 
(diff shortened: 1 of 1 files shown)


dzou:
Looks good, just 1 checkstyle issue remains: https://travis-ci.com/GoogleCloudPlatform/cloud-spanner-r2dbc/builds/137851138

PR #184, opened 2019-11-25
Use Google Cloud BOM instead of individual versions

elefeint:
Also removed commons-io and r2dbc-pool which are no longer used for testing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,7 +88,6 @@
     <reactor.version>Dysprosium-RELEASE</reactor.version>
 
     <google-auth.version>0.18.0</google-auth.version>
-    <grpc-bom.version>1.25.0</grpc-bom.version>
     <google-cloud-bom.version>2.9.0</google-cloud-bom.version>
 
     <mockito.version>3.1.0</mockito.version>
--- a/pom.xml
+++ b/pom.xml
@@ -83,17 +83,45 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
+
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -83,18 +83,46 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
+
     <r2dbc.version>0.8.0.RELEASE</r2dbc.version>
     <reactor.version>Dysprosium-RELEASE</reactor.version>
+
     <google-auth.version>0.18.0</google-auth.version>
+    <grpc-bom.version>1.25.0</grpc-bom.version>
+    <google-cloud-bom.version>0.119.0-alpha</google-cloud-bom.version>
```
meltsufin: Can you try using the `libraries-bom` instead? It should be also managing the grpc version.
https://github.com/GoogleCloudPlatform/cloud-opensource-java/wiki/The-Google-Cloud-Platform-Libraries-BOM

elefeint: That worked.

pom.xml
```diff
@@ -83,18 +83,38 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
+
     <r2dbc.version>0.8.0.RELEASE</r2dbc.version>
     <reactor.version>Dysprosium-RELEASE</reactor.version>
+
     <google-auth.version>0.18.0</google-auth.version>
+    <grpc-bom.version>1.25.0</grpc-bom.version>
```
meltsufin: unused now

PR #209, opened 2020-01-14
Performance microbenchmarks

### LINKED ISSUE #85: Performance testing
Comparison between using straight client library vs R2DBC driver implementation on the same data sets and number of requests.

elefeint:
Three types operations benchmarked: querying, DML and DDL.

Performing DDL with client library seems to have a lot of variability in performance. This may well be by design — GAX uses [exponential backoff](https://github.com/googleapis/gax-java/blob/a60bd347b69fbf725b0dd8ae18bbcf5b00b66b7b/gax-grpc/src/main/java/com/google/longrunning/stub/OperationsStubSettings.java#L242) with initial delay of 100ms to execute long-running operations. R2DBC by [default](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/blob/8eff19dbbac92fbbb42473795e6d810cc77565ce/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java#L155-L157) queries every 5 seconds up to a timeout of 10 minutes. I might at some point measure error behavior only.

For all other operations, the numbers look fairly similar between client library and R2DBC.

Full run output is in benchmark-run-20200114-1.txt and benchmark-run-20200114-2.txt.

Fixes #85 .


cloud-spanner-r2dbc-performance-testing/pom.xml
```diff
@@ -0,0 +1,196 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright (c) 2014, Oracle America, Inc.
```
dzou: remove copyright header for pom?

cloud-spanner-r2dbc-performance-testing/pom.xml
```diff
@@ -0,0 +1,196 @@
...
+            </plugin>
+        </plugins>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <artifactId>maven-clean-plugin</artifactId>
+                    <version>2.5</version>
+                </plugin>
+                <plugin>
+                    <artifactId>maven-deploy-plugin</artifactId>
```
dzou: Does this need to be configured so it does not get released?

cloud-spanner-r2dbc-performance-testing/src/main/java/com/google/cloud/spanner/r2dbc/benchmarks/BenchmarkState.java
```diff
@@ -0,0 +1,150 @@
...
+      this.dbClient = spanner.getDatabaseClient(DatabaseId.of(
+          options.getProjectId(), TEST_INSTANCE, TEST_DATABASE));
+
+      this.dbAdminClient = spanner.getDatabaseAdminClient();
+    }
+  }
+
+  @State(Scope.Benchmark)
+  public static class CommonState {
+    final Integer[] objectIds = new Integer[] {
```
dzou: Can you add comment describing what objectIds is?

cloud-spanner-r2dbc-performance-testing/pom.xml
elefeint: I'm going to find a lawyer for this one. It's out of my sphere of competence.

cloud-spanner-r2dbc-performance-testing/README.adoc
```diff
@@ -0,0 +1,75 @@
...
+Once the job is done, you will have a table named `met_objects` with 200,074 rows.
+
+If you are curious, this https://cloud.google.com/blog/products/gcp/when-art-meets-big-data-analyzing-200000-items-from-the-met-collection-in-bigquery[blog post] has more details about the data set.
+
+== Run benchmarks:
+
+```
+mvn clean install
+
+java -jar -Dspanner.instance=[INSTANCE] -Dspanner.database=[DATABASE] -Dgcp.project=[PROJECT] target/benchmarks.jar
```
meltsufin: Why aren't we using Maven here to run? Like `mvn:exec`.

cloud-spanner-r2dbc-performance-testing/README.adoc
```diff
@@ -0,0 +1,75 @@
+== Set up data
+
+These benchmarks use a public Met museum artwork data set in Cloud Spanner export format.
+The dataset is included on the top level of this repository.
+
+0. Unarchive `benchmark-data-met.tar.gz` into a folder `benchmark-data-met`.
```
meltsufin: Does this file exist somewhere that we can just link to it, rather than include it in the repo?
Also, please give the command to do this for convenience.

cloud-spanner-r2dbc-performance-testing/pom.xml
```diff
@@ -0,0 +1,196 @@
...
+        </dependency>
+    </dependencies>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+        <!--
+            JMH version to use with this project.
+          -->
+        <jmh.version>1.22</jmh.version>
```
meltsufin: Properties usually go at the top of the file.

cloud-spanner-r2dbc-performance-testing/pom.xml
```diff
@@ -0,0 +1,196 @@
...
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-spanner</artifactId>
+        </dependency>
+
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+            <scope>compile</scope>
```
meltsufin: You don't seem to be using `junit` and it almost never should be `compile` scope.

cloud-spanner-r2dbc-performance-testing/pom.xml
```diff
@@ -0,0 +1,196 @@
...
+                <version>3.8.0</version>
+                <configuration>
+                    <compilerVersion>${javac.target}</compilerVersion>
+                    <source>${javac.target}</source>
+                    <target>${javac.target}</target>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
```
meltsufin: This seems only necessary because you're not using Maven to run the project.

cloud-spanner-r2dbc-performance-testing/pom.xml
```diff
@@ -0,0 +1,196 @@
...
+                            </filters>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <artifactId>maven-clean-plugin</artifactId>
```
meltsufin: It's really odd to see all of these plugins version-managed. I would skip doing it unless there is a specific reason for it.

cloud-spanner-r2dbc-performance-testing/README.adoc
```diff
@@ -0,0 +1,75 @@
...
+
+```
+mvn clean install
+
+java -jar -Dspanner.instance=[INSTANCE] -Dspanner.database=[DATABASE] -Dgcp.project=[PROJECT] target/benchmarks.jar
+```
+
+To run individual benchmark class (in this case, `DmlBenchmark`):
+```
+java -jar -Dspanner.instance=[INSTANCE] -Dspanner.database=[DATABASE] -Dgcp.project=[PROJECT] target/benchmarks.jar DmlBenchmark
```
meltsufin: Can we have a parameter that controls the volume of the benchmark so that one can choose how long they want to run it for? In other words, if you want really accurate results, run for 30 minutes /1000 iterations, if you just want to get quick results after 2 minutes, pass some different parameters.

cloud-spanner-r2dbc-performance-testing/src/main/java/com/google/cloud/spanner/r2dbc/benchmarks/DdlBenchmark.java
```diff
@@ -0,0 +1,73 @@
...
+import org.openjdk.jmh.infra.Blackhole;
+import reactor.core.publisher.Flux;
+
+/**
+ * Benchmarks for DDL operations -- creating and dropping a table.
+ */
+public class DdlBenchmark extends BenchmarkState {
+
+  /**
+   * DDL with r2dbc driver.
```
meltsufin: Can we have more description in the javadoc explaining how many DDL statements and what kind are being executed?

cloud-spanner-r2dbc-performance-testing/src/main/java/com/google/cloud/spanner/r2dbc/benchmarks/DmlBenchmark.java
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * DML with r2dbc driver.
+   */
+  @Benchmark
+  public Integer testDmlR2dbcDriver(R2dbcConnectionState r2dbcState, CommonState common) {
+
+    String query = common.getSingleRowUpdateQuery();
+
+    Integer result =
+        Flux.from(r2dbcState.r2dbcConnection.createStatement(query).execute())
```
meltsufin: Are you starting a read-write transaction here as well?

cloud-spanner-r2dbc-performance-testing/src/main/java/com/google/cloud/spanner/r2dbc/benchmarks/QueryingBenchmark.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private static final String QUERY
+      = "SELECT * FROM met_objects LIMIT ";
+
+  /**
+   * querying with r2dbc driver.
+   */
+  @Benchmark
+  public List<String> testQueryingR2dbc(R2dbcConnectionState r2dbcState) {
+    int numRows = 10;
```
meltsufin: Can this be a function argument?

pom.xml
```diff
@@ -329,6 +325,13 @@
       </modules>
     </profile>
 
+    <profile>
+      <id>performance-testing</id>
```
meltsufin: What's the purpose of a separate profile here, rather than the default?

cloud-spanner-r2dbc-performance-testing/README.adoc
elefeint: Not in this format, it does not. I could make a public storage bucket, but then someone needs to remember it exists and not delete it?

cloud-spanner-r2dbc-performance-testing/pom.xml
elefeint: blame the archetype. But yeah, I'll move it.

cloud-spanner-r2dbc-performance-testing/pom.xml
elefeint: this is a test-only project, but yeah. (continues blaming archetype)

cloud-spanner-r2dbc-performance-testing/README.adoc
elefeint: those parameters come free with JMH actually. I'll document.

cloud-spanner-r2dbc-performance-testing/src/main/java/com/google/cloud/spanner/r2dbc/benchmarks/DmlBenchmark.java
elefeint: nope. Was going to add transaction testing separately; this PR is a bit big already and I wanted to shove it out of my brain.

cloud-spanner-r2dbc-performance-testing/src/main/java/com/google/cloud/spanner/r2dbc/benchmarks/QueryingBenchmark.java
elefeint: Definitely. I was going to parameterize but forgot.

pom.xml
elefeint: to avoid building it on PRs mostly. Since we won't run benchmarks, no point losing the time.

cloud-spanner-r2dbc-performance-testing/README.adoc
meltsufin: Ok, in that case you can just leave it in the repo. It's not a huge file.

cloud-spanner-r2dbc-performance-testing/src/main/java/com/google/cloud/spanner/r2dbc/benchmarks/DmlBenchmark.java
meltsufin: But in the client lib test for the same thing, it looks like you are starting a transaction. So, not apples to apples.

pom.xml
meltsufin: Building won't run them automatically presumably but detect compile errors. If we want to detect performance regressions, we may actually want to run the benchmarks as part of our CI.

cloud-spanner-r2dbc-performance-testing/pom.xml
elefeint: Header comment stays. I'll address the remaining feedback in the file.

cloud-spanner-r2dbc-performance-testing/README.adoc
```diff
@@ -0,0 +1,77 @@
+== Set up data
+
+These benchmarks use a public Met museum artwork data set in Cloud Spanner export format.
```
dmitry-s: maybe add a link to the data source?

dmitry-s: even better, instead of the archive file just document how to download the data set from the source repo

elefeint: There is a link below, with "If you are curious". But I can't provide a data set in Spanner export format, unless we are willing to host a public storage bucket.
I've converted this from a BigQuery data set into Spanner a while ago; I don't think this exists anywhere else.

cloud-spanner-r2dbc-performance-testing/README.adoc
elefeint: It's possible, but it would require adding a main() method to have a runnable entry point and adding all benchmark classes to the configuration. Then if we want ability to exclude/include individual classes, we'd have to set up custom parameters that main method would pass into JMH runner.

Too much hassle given that realistically we are neither going to ever run those from an IDE nor make them a part of the build.

cloud-spanner-r2dbc-performance-testing/pom.xml
elefeint: JMH archetype generates this; if we want to have benchmarks running automatically, we can create our own configuration with programmatic JMH invocation.

cloud-spanner-r2dbc-performance-testing/pom.xml
elefeint: This won't be part of the release maven profile; on the other hand, can't hurt.

cloud-spanner-r2dbc-performance-testing/src/main/java/com/google/cloud/spanner/r2dbc/benchmarks/DmlBenchmark.java
elefeint: Every standalone DML R2DBC statement runs in a read-write transaction that we create as part of sending a query, and then commit [1], so it's similar.

[1] https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/blob/572212ca74d827cdace1a6af91e0ccce42d80fba/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/client/GrpcClient.java#L239

pom.xml
elefeint: Possibly, but not these benchmarks -- they are explicitly for comparing client library to R2DBC. 
We could add benchmarks and send results to stackdriver in integration tests -- these run on every commit, so would give us a good view into any regressions.

elefeint:
Not planning to merge this into the repo.

PR #222, opened 2020-02-13
Add milestones repo to the sample

elefeint:
Spring Boot autoconfiguration for R2DBC is still at milestone level, even in experimental.

This should fix the build.


elefeint:
Spring Boot autoconfiguration for R2DBC is still at milestone level.

This should fix the build.

PR #262, opened 2020-08-21
Client library-based driver implementation

elefeint:
This PR introduces a parallel implementation of the driver based on the client library, with the goal of providing consistent behavior for JDBC and R2DBC Cloud Spanner drivers. 

For now, gRPC implementation remains primary; switching to the client library implementation requires providing `.option(Option.valueOf("client-implementation"), "client-library")` when creating a connection factory.

* Query with parameter binding
* DML with transactions
* DDL temporarily falls back on gRPC implementation.


elefeint:
Oh, many times over -- it was in PRs merged against the async branch. It's finally all coming home.

PR #263, opened 2020-08-25
Customizable executor thread count 

### LINKED ISSUE #247: Allow customizing callback executor thread count
Allow executor customization (number of threads) through connection properties.

Depends on #246.

dmitry-s:
fixes #247

### DIFF
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
@@ -53,7 +53,7 @@ public class SpannerConnectionConfiguration {
 
   private Duration ddlOperationPollInterval;
 
-  private int executorThreads;
+  private int threadPoolSize;
 
   /**
    * Constructor which initializes the configuration from an Cloud Spanner R2DBC url.
@@ -137,8 +137,8 @@ public class SpannerConnectionConfiguration {
     return this.ddlOperationPollInterval;
   }
 
-  public int getExecutorThreads() {
-    return this.executorThreads;
+  public int getThreadPoolSize() {
+    return this.threadPoolSize;
   }
 
   @Override
@@ -185,7 +185,7 @@ public class SpannerConnectionConfiguration {
 
     private Duration ddlOperationPollInterval = Duration.ofSeconds(5);
 
-    private int executorThreads = Runtime.getRuntime().availableProcessors();
+    private Integer threadPoolSize;
 
     public Builder setUrl(String url) {
       this.url = url;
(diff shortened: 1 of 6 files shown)


cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -53,6 +53,8 @@
 
   private Duration ddlOperationPollInterval;
 
+  private int executorThreads;
```
meltsufin, line 56: `threadPoolSize` maybe a better name?

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfigurationTest.java
```diff
@@ -116,6 +116,28 @@ public void partialResultSetFetchSize() {
     assertThat(config.getPartialResultSetFetchSize()).isEqualTo(42);
   }
 
+  @Test
+  public void partialResultSetExecutorThreads() {
```
elefeint, line 120: Remove `partialResultSet` prefix in this and the below?

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfigurationTest.java
```diff
@@ -116,6 +116,28 @@ public void partialResultSetFetchSize() {
...
+    SpannerConnectionConfiguration config = this.configurationBuilder
+        .setExecutorThreads(16)
```
elefeint, line 122: Make it a weirder number? This is too plausible as a number of processors, so could hypothetically pass even if the custom number is not set.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -179,6 +185,8 @@ public int hashCode() {
 
     private Duration ddlOperationPollInterval = Duration.ofSeconds(5);
 
+    private int executorThreads = Runtime.getRuntime().availableProcessors();
```
elefeint, line 188: It would be good to only access system environment lazily -- only if number of processors has not been set. I would move the initialization into `build()` method, so that if it's 0 (which is clearly invalid), only then the `Runtime` gets checked.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
```diff
@@ -52,6 +52,8 @@
    */
   public static final Option<String> URL = Option.valueOf("url");
 
+  public static final Option<Integer> POOL_SIZE = Option.valueOf("thread_pool_size");
```
elefeint, line 55: nit: add "THREAD_" prefix to the constant, just because it's a public constant.

PR #264, opened 2020-08-31
Run SELECT statements in transactions

### LINKED ISSUE #242: Make basic read and write transactions work
with client library

### LINKED ISSUE #253: Test interspersing DML with SELECT statement
Test that having SELECT statements in a write transaction reads correct state.
Add integration test.

### LINKED ISSUE #254: DML parameter binding
Change SpannerClientLibraryDmlStatement to Implement parameter binding.

elefeint:
Fixes #253 
Fixes #242.

Fixes #254, since I refactored Select/DML statements to inherit from the same `AbstractSpannerClientLibraryStatement` (happy to rename it more meaningfully to `ParameterBindingClientLibraryStatement`).

Other changes:
* Renamed some `DatabaseClientReactiveAdapter` state fields for readability: currentTransactionFuture -> txnContext, asyncTransactionLastStep -> lastStep.
* I changed the future-to-publisher conversion helpers to return void and to send data to a passed-in publisher sink because helper methods returning `ApiFuture` proved impossible to adapt for SELECT statements: `.then()` expects an ApiFuture, but all SELECT functionality returns `AsyncResultSet`. So there was a conflict between the type that can be converted to a Flux (`AsyncResultSet`) and the type that the client library transaction chaining requires (`ApiFuture`). Passing a sink into all of that conversion instead of creating it externally from a nicely packaged `ApiFuture` solves all of this. I renamed `convertFutureToMono` to `adaptFutureToMono` to reflect the change. This is all private helper methods, so not public API at all.

The lambda formatting is a grudging compromise between me and checkstyle. Suggestions on making it look better gleefully accepted.

### DIFF
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
@@ -139,11 +139,9 @@ class DatabaseClientReactiveAdapter {
    * @return reactive pipeline for closing the connection.
    */
   public Mono<Void> close() {
-    return Mono.<Void>fromSupplier(
-        () -> {
-          this.transactionManager.close();
-          return null;
-        });
+    // TODO: if txn is committed/rolled back and then connection closed, clearTransactionManager
+    // will run twice, causing trace span to be closed twice. Introduce `closed` field.
+    return Mono.<Void>fromRunnable(this::clearTransactionManager);
   }
 
   /**
@@ -191,7 +189,7 @@ class DatabaseClientReactiveAdapter {
         sink -> {
           if (this.isInTransaction()) {
 
-            LOGGER.debug("  chainingSELECT statement in transaction: " + statement.getSql());
+            LOGGER.debug("  chaining SELECT statement in transaction: " + statement.getSql());
             // The first statement in a transaction has no input, hence Void input type.
             // The subsequent statements take the previous statement's return value as input.
             this.lastStep =
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
@@ -19,8 +19,6 @@ package com.google.cloud.spanner.r2dbc.v2;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutureCallback;
 import com.google.api.core.ApiFutures;
-import com.google.cloud.ByteArray;
-import com.google.cloud.Date;
 import com.google.cloud.Timestamp;
 import com.google.cloud.spanner.AsyncResultSet;
 import com.google.cloud.spanner.AsyncResultSet.CallbackResponse;
@@ -28,16 +26,8 @@ import com.google.cloud.spanner.AsyncTransactionManager;
 import com.google.cloud.spanner.AsyncTransactionManager.AsyncTransactionStep;
 import com.google.cloud.spanner.AsyncTransactionManager.TransactionContextFuture;
 import com.google.cloud.spanner.DatabaseClient;
-import com.google.cloud.spanner.SpannerException;
 import com.google.cloud.spanner.Statement;
-import com.google.cloud.spanner.Struct;
-import com.google.cloud.spanner.StructReader;
-import com.google.cloud.spanner.Type;
-import com.google.common.base.Function;
-import com.google.common.collect.ImmutableList;
-import com.google.spanner.v1.ResultSetStats;
-import java.util.List;
-import java.util.concurrent.Executor;
+import com.google.cloud.spanner.TransactionContext;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Supplier;
 import org.reactivestreams.Publisher;
@@ -46,6 +36,7 @@ import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.FluxSink;
 import reactor.core.publisher.Mono;
+import reactor.core.publisher.MonoSink;
 
 /**
  * Converts gRPC/Cloud Spanner client library asyncronous abstractions into reactive ones.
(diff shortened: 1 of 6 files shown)


cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
```diff
@@ -109,82 +118,143 @@ private void clearTransactionManager() {
...
-          }, this.executorService);
-    });
+          } catch (Throwable t) {
+            sink.error(t);
+            return CallbackResponse.DONE;
+          }
+        });
   }
 
+  private <T> void adaptFutureToMono(ApiFuture<T> future, MonoSink<T> sink) {
```
dzou, line 242: Minor refactor suggestion - For the Mono case I would prefer keeping it `convertFutureToMono` and returning a Mono if possible.

I feel it's easy to forget to call the `adaptFutureToMono` method where you need it. If you have more complex if-branches, like in the `runDmlStatement` you need to remember to call it in both the `if` and `else`. We might make a mistake if there is more complex branching.

I just prefer the return method way since you'll get a guaranteed missing return statement compile error with this refactoring.

For example:

new helper:

```
  private <T> Mono<T> convertFutureToMono(ApiFuture<T> future) {
    return Mono.create(sink -> {
      sink.onCancel(() -> future.cancel(true));
      ApiFutures.addCallback(
          future,
          new ApiFutureCallback<T>() {
            @Override
            public void onFailure(Throwable t) {
              sink.error(t);
            }

            @Override
            public void onSuccess(T result) {
              sink.success(result);
            }
          },
          this.executorService);
    });
  }
```

Then begin transaction might look like this with `Mono.defer` (which is what we were using in the non-client library Spanner R2DBC). You'll get a missing return statement error if you forget the `return convertFutureToMono(this.txnContext)`.

```
  public Mono<Void> beginTransaction() {
    return Mono.defer(() -> {
      LOGGER.debug("begin transaction");
      this.transactionManager = this.dbClient.transactionManagerAsync();
      this.txnContext = this.transactionManager.beginAsync();
      return convertFutureToMono(this.txnContext);
    }).then();
  }
```

But ya - these comments don't apply to your `Flux` helper method since it needs to accept the sink for the select statement. But if possible I would try to minimize the number of methods that accept and mutate `sinks`; i only fear those bugs where we called `sink.complete()` too early and then try to call `sink.next(..)` after etc.

skuruppu:
CC @olavloite

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
elefeint: Yep, `convertFutureToMono()` ends up being a bit less verbose and safer.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
```diff
@@ -110,81 +115,145 @@ private void clearTransactionManager() {
...
+          }
+        });
+  }
 
-    return this.asyncTransactionLastStep;
+  private ApiFuture<Void> runSelectStatement(
+      Supplier<ReadContext> ctxSupplier,
+      com.google.cloud.spanner.Statement statement,
+      FluxSink<SpannerClientLibraryRow> sink) {
+    AsyncResultSet ars = ctxSupplier.get().executeQueryAsync(statement);
```
olavloite, line 211: This `get()` could be blocking in case it is the first statement of the transaction and the `ApiFuture` that was returned by `beginAsync()` is not done. I don't know if that's a problem here?

elefeint: When this helper method is called from a transaction context (where `beginTransaction()` was called explicitly, it is chained to either [`AsyncTransactionContext`](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/264/files/ea4e9cdba1d4db9ca250ed28bfe01f8b1944fe8c#diff-3c2541302ebdc09547a981a571dd5ffcR160) or to the[ previous step](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/264/files/ea4e9cdba1d4db9ca250ed28bfe01f8b1944fe8c#diff-3c2541302ebdc09547a981a571dd5ffcR195). At that point, it's guaranteed to run after the previous future resolves, right?

I am actually more concerned about this method being called without an explicit transaction, because then `DatabaseClient.singleUse()` [is called](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/264/files/ea4e9cdba1d4db9ca250ed28bfe01f8b1944fe8c#diff-3c2541302ebdc09547a981a571dd5ffcR202), and I recall there used to be blocking when getting a session for the first time. Is there still?

Theoretical aside: R2DBC specification says "A Connection object is not safe for concurrent state-changing by multiple subscribers. A connection object can be shared across multiple threads that serially run operations by using appropriate synchronization mechanisms", so it's generally safe to assume that beginTransaction-runStatement-commitTransaction are done sequentially.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
```diff
@@ -110,81 +115,145 @@ private void clearTransactionManager() {
...
+        // The first statement in a transaction has no input, hence Void input type.
+        // The subsequent statements take the previous statements' return (affected row count)
+        // as input.
+        AsyncTransactionStep<? extends Object, Long> updateStatementFuture =
+            this.lastStep == null
+                ? this.txnContext.then(
+                    (ctx, unusedVoid) -> ctx.executeUpdateAsync(statement),
+                    this.executorService)
+                : this.lastStep.then(
+                    (ctx, previousRowCount) -> ctx.executeUpdateAsync(statement),
```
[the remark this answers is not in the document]
elefeint: You are right; I'll fix it -- I missed this from when there was only DML implemented.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
olavloite: > When this helper method is called from a transaction context (where beginTransaction() was called explicitly, it is chained to either AsyncTransactionContext or to the previous step. At that point, it's guaranteed to run after the previous future resolves, right?

Yes, correct. So that should not be a problem if it is only used in that way.

> I am actually more concerned about this method being called without an explicit transaction, because then DatabaseClient.singleUse() is called, and I recall there used to be blocking when getting a session for the first time. Is there still?

No, getting a session is now non-blocking. So `DatabaseClient#singleUse()` is guaranteed to be non-blocking.

meltsufin (dismissed):
LGTM

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
elefeint: I added documentation to the method to make sure it's not misused in the future.

dzou (approved):

meltsufin (approved):

PR #272, opened 2020-09-24
Add support for binding Java integers as Spanner parameters

elefeint:
Even with the understanding that Spanner only has INT64, we can broaden Integer into Long automatically when binding to simplify user experience. Binding integers was supported in V1.

### DIFF
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryBinder.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryBinder.java
@@ -32,7 +32,8 @@ class ClientLibraryBinder {
   private static List<ClientLibraryTypeBinder> buildBinders() {
     List<ClientLibraryTypeBinder> binders = new ArrayList<>();
     binders.add(
-        new ClientLibraryTypeBinderImpl(Integer.class, (binder, val) -> binder.to(Long.valueOf( (Integer) val) )));
+        new ClientLibraryTypeBinderImpl(Integer.class,
+            (binder, val) -> binder.to(longFromInteger((Integer) val))));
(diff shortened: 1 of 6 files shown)


cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryBinder.java
```diff
@@ -31,6 +31,9 @@
 
   private static List<ClientLibraryTypeBinder> buildBinders() {
     List<ClientLibraryTypeBinder> binders = new ArrayList<>();
+    binders.add(
+        new ClientLibraryTypeBinderImpl(Integer.class,
+            (binder, val) -> binder.to(longFromInteger((Integer) val))));
```
meltsufin, line 36: Can `ClientLibraryTypeBinder` be parameterized so that you don't have to cast to `Integer`?

elefeint: It can, but that's just going to change from this:
```
new ClientLibraryTypeBinderImpl(Long.class, (binder, val) -> binder.to((Long) val))
```
to this:
```
new ClientLibraryTypeBinderImpl<Long>(Long.class, (binder, val) -> binder.to(val))
```

I don't have a strong opinion about which is more readable. Thoughts?

elefeint: Actually it can become `new ClientLibraryTypeBinderImpl<>(Long.class, (binder, val) -> binder.to(val))`, which is a little nicer.

dmitry-s (approved):

meltsufin (approved):
Nice!

PR #278, opened 2020-10-01
Bind multiple statements

elefeint:
This PR covers the `Statement.bind().bind().add()` flow for both SELECT and DML.

Other changes:
* `AbstractSpannerClientLibraryStatement`
  * Reduced visibility of most state to private.
  * Implemented `execute` to complete the final row if necessary, and delegate actual query execution to abstract templated methods  `executeSingle()` and `executeMultiple()`.
* Added `runBatchDml()` method delegating to the client library's `TransactionContext.batchUpdateAsync()`.
   * Refactored most of `runDmlStatement()` into a helper method, to be reusable by `runBatchDml().
* Re-enabled two TCK tests that needed the multi-bind functionality.
* Added many tests.

### DIFF
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
@@ -41,7 +41,7 @@ abstract class AbstractSpannerClientLibraryStatement implements Statement {
 
   private com.google.cloud.spanner.Statement.Builder currentStatementBuilder;
 
-  private boolean incompleteBinding = false;
+  private boolean startedBinding = false;
 
   private List<com.google.cloud.spanner.Statement> statements;
 
@@ -67,7 +67,7 @@ abstract class AbstractSpannerClientLibraryStatement implements Statement {
 
     this.statements.add(this.currentStatementBuilder.build());
     this.currentStatementBuilder = com.google.cloud.spanner.Statement.newBuilder(this.query);
-    this.incompleteBinding = false;
+    this.startedBinding = false;
 
     return this;
   }
@@ -75,9 +75,9 @@ abstract class AbstractSpannerClientLibraryStatement implements Statement {
   @Override
   public Publisher<? extends Result> execute() {
     if (this.statements != null) {
-      if (this.incompleteBinding) {
+      if (this.startedBinding) {
         this.statements.add(this.currentStatementBuilder.build());
-        this.incompleteBinding = false;
+        this.startedBinding = false;
         this.currentStatementBuilder = null;
       }
       return executeMultiple(this.statements);
@@ -99,7 +99,7 @@ abstract class AbstractSpannerClientLibraryStatement implements Statement {
   @Override
   public Statement bind(String name, Object value) {
     ClientLibraryBinder.bind(this.currentStatementBuilder, name, value);
-    this.incompleteBinding = true;
+    this.startedBinding = true;
     return this;
   }
 
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
@@ -39,7 +39,7 @@ abstract class AbstractSpannerClientLibraryStatement implements Statement {
 
   protected final DatabaseClientReactiveAdapter clientLibraryAdapter;
 
-  protected com.google.cloud.spanner.Statement.Builder currentStatementBuilder;
+  private com.google.cloud.spanner.Statement.Builder currentStatementBuilder;
 
   private boolean incompleteBinding = false;
 
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
@@ -85,10 +85,10 @@ abstract class AbstractSpannerClientLibraryStatement implements Statement {
     return executeSingle(this.currentStatementBuilder.build());
   }
 
-  public abstract Mono<SpannerClientLibraryResult> executeSingle(
+  protected abstract Mono<SpannerClientLibraryResult> executeSingle(
       com.google.cloud.spanner.Statement statement);
 
-  public abstract Flux<SpannerClientLibraryResult> executeMultiple(
+  protected abstract Flux<SpannerClientLibraryResult> executeMultiple(
       List<com.google.cloud.spanner.Statement> statements);
 
   @Override
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
@@ -24,6 +24,8 @@ import java.util.List;
 import org.reactivestreams.Publisher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 /**
  * Cloud Spanner base implementation of R2DBC SPI for query and DML statements.
@@ -37,10 +39,13 @@ abstract class AbstractSpannerClientLibraryStatement implements Statement {
 
   protected final DatabaseClientReactiveAdapter clientLibraryAdapter;
 
-  protected final com.google.cloud.spanner.Statement.Builder currentStatementBuilder;
+  protected com.google.cloud.spanner.Statement.Builder currentStatementBuilder;
+
+  private boolean incompleteBinding = false;
 
   private List<com.google.cloud.spanner.Statement> statements;
 
+  private final String query;
 
   /**
    * Creates a ready-to-run Cloud Spanner statement.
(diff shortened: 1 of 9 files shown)


cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
```diff
@@ -34,7 +39,13 @@
 
   protected final DatabaseClientReactiveAdapter clientLibraryAdapter;
 
-  protected final Builder statementBuilder;
+  private com.google.cloud.spanner.Statement.Builder currentStatementBuilder;
+
+  private boolean incompleteBinding = false;
```
dzou, line 44: I would rename to `startedBinding`; think this may be more clear to express the boolean

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
```diff
@@ -44,23 +55,51 @@
...
     this.clientLibraryAdapter = clientLibraryAdapter;
-    this.statementBuilder = com.google.cloud.spanner.Statement.newBuilder(query);
+    this.currentStatementBuilder = com.google.cloud.spanner.Statement.newBuilder(query);
+    this.query = query;
   }
 
   @Override
   public Statement add() {
-    throw new UnsupportedOperationException();
+    if (this.statements == null) {
```
dzou, line 64: Is this class meant to be re-used after a user calls `execute`? If not, I would just initialize the `this.statements` in the constructor once.

But if it is meant to be re-used, I think it may make sense to keep all the state resetting logic inside `execute()` just to keep in once place. `execute()` can be the method which runs the statements and then clears the list.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryStatement.java
```diff
@@ -41,10 +42,17 @@ public SpannerClientLibraryStatement(
...
     return this.clientLibraryAdapter
-        .runSelectStatement(this.statementBuilder.build())
-        .transform(rows -> Mono.just(new SpannerClientLibraryResult(rows, Mono.empty())));
+        .runSelectStatement(statement)
+        .transform(rows -> Mono.just(new SpannerClientLibraryResult(rows, Mono.empty()))).single();
   }
 
+  @Override
+  protected Flux<SpannerClientLibraryResult> executeMultiple(List<Statement> statements) {
+    return Flux.fromIterable(statements).flatMapSequential(statement ->
```
dzou, line 53: Not 100% sure on this; but I think you can just delegate a call to `executeSingle(...)` via:

`return Flux.fromIterable(statements).flatMapSequential(statement -> executeSingle(statement));`

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
ttomsu: Is the goal to make this multi-threaded safe? Because I don't know if this state variable would survive well in a race condition.

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/it/ClientLibraryBasedIT.java
```diff
@@ -380,6 +400,163 @@ public void ddlCreateAndDrop() {
...
+                    .bind("words", 0).add()
+                    .execute()
+            )
+            .flatMapSequential(rs -> rs.map((row, rmeta) -> row.get(1, Long.class))))
+    ).expectNext(0L, 0L, 1L, 2L, 3L).as("Row count matches bound variables").verifyComplete();
+
+  }
+
+  /* This test
+  1) exercises s a different internal code path than selectMultipleBoundParameterSetsNoTransaction()
```
ttomsu, line 444: stray 's'

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
elefeint: For Connection, the spec explicitly says that "connection object can be shared across multiple threads that serially run operations by using appropriate synchronization mechanisms" -- so it's expected to be externally synchronized and does not need to take care of simultaneously modified state. For Statement, there is no explicit clause like that, but looking at the reference H2 implementation, the same assumption is made: https://github.com/r2dbc/r2dbc-h2/blob/main/src/main/java/io/r2dbc/h2/H2Statement.java

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
elefeint: The spec says "Parameterized statements may be cached by R2DBC implementations for reuse (for example, for prepared statement execution).", so it's left up to implementation. I would prefer to at least not preclude reuse of a `Statement`, even if we don't explicitly support it.

Which state resetting do you mean? The typical flow is:
```
conn.createStatement(statement)
                    .bind("uuid", uuid1).bind("title", "A").add()
                    .bind("uuid", uuid2).bind("title", "B").add()
                    .bind("uuid", uuid3).bind("title", "C").add()
                    .execute()
```

So the in-progress binding needs to be reset after each `add()` (so, theoretically, could be done in `bind()` with some logic), but it's not enough to reset it in `execute()`.

Unless you meant that bound variables should be removed after `execute()`? I don't think so -- if we are going to allow reuse, the value is in keeping the statement with the parameters bound. Creating a new Statement is a simple one liner, but binding variables could be involved.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryStatement.java
elefeint: Good catch, thanks!

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/it/ClientLibraryBasedIT.java
elefeint: I am just going to blame the cat.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
dzou: Ah I see, that makes sense.

What I meant with the state-clearing was that `execute()` would need to clear the statements list I think if you want to re-use it (either set to null or `.clear()`), but I don't think this is being done yet.

I think it might be simpler to just do `this.statements = new ArrayList<>();` in the constructor, and then do a `this.statements.clear()` in the `execute()` method after running; then you can remove all the null-checks I believe.

dzou (dismissed):

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
```diff
@@ -44,23 +55,51 @@
...
+    this.statements.add(this.currentStatementBuilder.build());
+    this.currentStatementBuilder = com.google.cloud.spanner.Statement.newBuilder(this.query);
+    this.startedBinding = false;
+
+    return this;
   }
 
+  @Override
+  public Publisher<? extends Result> execute() {
+    if (this.statements != null) {
```
meltsufin, line 77: Like Dan suggested, it might be easier to just make sure that `this.statements` is never `null`.
You can even re-use it for single statements where you would check for `size() == 0`.

meltsufin (approved):
Just one optional suggestion. Otherwise looks good to merge!

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
elefeint: Right, I was trying to avoid creating a list by default because the most common scenario is to have a string query with only a single set of bound parameters. Since statement execution is the most common operation for a database, it may make sense to save on an extra object per statement.

meltsufin: OK, I see.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
```diff
@@ -166,31 +170,44 @@ private void clearTransactionManager() {
    * @return reactive pipeline for running a DML statement
    */
   public Mono<Long> runDmlStatement(com.google.cloud.spanner.Statement statement) {
+    return runBatchDmlInternal(ctx -> ctx.executeUpdateAsync(statement));
+  }
+
+  /**
+   * Allows running DML statements in a batch.
+   *
+   * <p>If no transaction is active, a single-use transaction will be used.
```
dmitry-s, line 179:
```suggestion
   * Allows running DML statements in a batch.
   * <p>
   * If no transaction is active, a single-use transaction will be used.
```

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryDmlStatement.java
```diff
@@ -42,13 +43,25 @@ public SpannerClientLibraryDmlStatement(DatabaseClientReactiveAdapter clientLibr
...
 
+  @Override
+  protected Flux<SpannerClientLibraryResult> executeMultiple(List<Statement> statements) {
+    return this.clientLibraryAdapter
+        .runBatchDml(statements)
+        .flatMapMany(numRowsArray ->
+              Flux.fromArray(ArrayUtils.toObject(numRowsArray))
+                  .map(numRows ->
+                      new SpannerClientLibraryResult(Flux.empty(), Mono.just(longToInt(numRows))))
+
```
dmitry-s, line 61:
```suggestion
```

dmitry-s (changes requested):

meltsufin (approved):

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
elefeint: google checkstyle is yelling at me now -- "tag should be placed immediately before the first word, with no space after." :(

meltsufin (approved):

PR #286, opened 2020-11-03
support Long to Integer conversion on read

### LINKED ISSUE #276: Determine correct behavior for Row.get("column", Integer.class)
Cloud Spanner only has INT64, a long. The current V2 behavior is to ignore the requested type and instead only use the type of the retrieved column to decode the value.

In V1, retrieving an Integer column value was supported.

Should we support retrieving integers if the user requested an Integer explicitly? Likely yes, with exceptions thrown if the returned INT64 value does not fit into an integer. The alternative would be to disallow this scenario, but that's a regression from V1.

dmitry-s:
fixes #276

### DIFF
--- a/cloud-spanner-r2dbc/pom.xml
+++ b/cloud-spanner-r2dbc/pom.xml
@@ -5,6 +5,18 @@
 
   <artifactId>cloud-spanner-r2dbc</artifactId>
   <packaging>jar</packaging>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>11</source>
+          <target>11</target>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 
   <parent>
     <groupId>com.google.cloud</groupId>
--- a/cloud-spanner-r2dbc/pom.xml
+++ b/cloud-spanner-r2dbc/pom.xml
@@ -5,18 +5,6 @@
 
   <artifactId>cloud-spanner-r2dbc</artifactId>
   <packaging>jar</packaging>
-  <build>
-    <plugins>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-compiler-plugin</artifactId>
-        <configuration>
-          <source>11</source>
-          <target>11</target>
-        </configuration>
-      </plugin>
-    </plugins>
-  </build>
 
   <parent>
     <groupId>com.google.cloud</groupId>
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryDecoder.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryDecoder.java
@@ -95,6 +95,6 @@ public class ClientLibraryDecoder {
     if (type.isAssignableFrom(value.getClass())) {
       return (T) value;
     }
-    return SpannerClientLibrariesConverters.convert(value, type);
+    return SpannerClientLibraryConverters.convert(value, type);
   }
 }
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryDecoder.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryDecoder.java
@@ -84,8 +84,17 @@ public class ClientLibraryDecoder {
     Object value =
         struct.isNull(index)
             ? null
-            : selectedCodecsMap.get(struct.getColumnType(index)).apply(struct, index);
+            : readAndConvert(struct, index, selectedCodecsMap, type);
 
     return (T) value;
+  }
+
+  private static <T> T readAndConvert(Struct struct, int index,
+      Map<Type, BiFunction<Struct, Integer, Object>> selectedCodecsMap, Class<T> type) {
+    Object value = selectedCodecsMap.get(struct.getColumnType(index)).apply(struct, index);
+    if (type.isAssignableFrom(value.getClass())) {
+      return (T) value;
+    }
+    return SpannerClientLibrariesConverters.convert(value, type);
   }
 }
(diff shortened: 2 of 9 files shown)


cloud-spanner-r2dbc/pom.xml
```diff
@@ -5,6 +5,18 @@
 
   <artifactId>cloud-spanner-r2dbc</artifactId>
   <packaging>jar</packaging>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>11</source>
```
elefeint, line 14: Java 8 came from these properties below, which I think are read by the same plugin, so it's easier to change them there:
```    
```

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibrariesConverters.java
```diff
@@ -0,0 +1,45 @@
...
+import com.google.cloud.spanner.r2dbc.v2.conversion.SpannerClientLibrariesConverter;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class SpannerClientLibrariesConverters {
+
+  private static final List<SpannerClientLibrariesConverter<?>> converters = createConverters();
+
+  private static ArrayList<SpannerClientLibrariesConverter<?>> createConverters() {
```
elefeint, line 29: Return generic `List`?

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryDecoder.java
```diff
@@ -84,8 +84,17 @@
...
     return (T) value;
   }
+
+  private static <T> T readAndConvert(Struct struct, int index,
+      Map<Type, BiFunction<Struct, Integer, Object>> selectedCodecsMap, Class<T> type) {
+    Object value = selectedCodecsMap.get(struct.getColumnType(index)).apply(struct, index);
+    if (type.isAssignableFrom(value.getClass())) {
+      return (T) value;
+    }
+    return SpannerClientLibrariesConverters.convert(value, type);
```
elefeint, line 98: Having it be a static method makes it not that easy to unit test separately from the specific converter logic. What do you think about making `SpannerClientLibrariesConverters` have state?

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryDecoderTest.java
```diff
@@ -38,21 +39,36 @@
...
+    );
+  }
+
   static Stream<Arguments> data() {
     return Stream.of(
         // Arrays
         arguments(
             long[].class,
-            new long[] {123456L, 45678L},
+            new long[]{123456L, 45678L},
```
elefeint, line 63: Checkstyle made you do this?! Sounds unlike the usual style.

elefeint (changes requested):
I like the `SpannerClientLibrariesConverter` abstraction. Great tests, too.

You may be able to remove a test override `SpannerClientLibraryTestKit.transactionCommit()`, and its helper `extractColumnsLong()`.

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryDecoderTest.java
dmitry-s: GoogleStyle plugin did that.

cloud-spanner-r2dbc/pom.xml
elefeint: Also, .travis.yml needs to change to run on the newer JDK - - right now, the build fails with `Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project cloud-spanner-r2dbc: Fatal error compiling: invalid target release: 11 -> [Help 1]`

elefeint (commented):
It may be easier to reason about PRs if the JDK changes are in their own tiny standalone one.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryDecoder.java
dmitry-s: How about we keep it like this for now and if/when we need we'll switch to stateful `SpannerClientLibrariesConverters`?

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibrariesConverters.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.google.cloud.spanner.r2dbc.v2;
+
+import com.google.cloud.spanner.r2dbc.v2.conversion.LongIntegerConverter;
+import com.google.cloud.spanner.r2dbc.v2.conversion.SpannerClientLibrariesConverter;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class SpannerClientLibrariesConverters {
```
meltsufin, line 25: Javadoc? The name sounds odd. Did you mean `SpannerClientLibraryConverters`?
Also, instead of making it a collection of `static` methods, why not just make it the converter?

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/ClientLibraryDecoder.java
meltsufin: Can you have an interface and implementation that has a single method `T convert(Struct struct, int index, Class<T> type)`? Then all of these conversion supporting decoders can move in there. I think it would be much easier to test conversion logic that way.

meltsufin (commented):
I think there are opportunities here to improve the conversion design.
However, if you think that would be better done in a separate PR, that's fine too.

elefeint (approved):
Let's get it in; we can always decide on a refactoring later.

meltsufin:
@dmitry-s Do you want to merge this in?

meltsufin (approved):

PR #287, opened 2020-11-03
V2 extended transactions

### LINKED ISSUE #249: Allow customization of transaction types (timestamp bound vs regular)
Until extensible transaction definitions are available, handle it the same way as V1 -- allow downcasting connection and getting a custom `beginTransactions()` method.

elefeint:
Read/write and readonly transactions use completely different concepts in client library (`TransactionManager` is only for read/write transactions), so I created `DatabaseClientTransactionManager` to encapsulate all the random bits of transactional state.

All forms of stale readonly transactions are supported with this PR.

Partitioned DML is out of scope, since it does not fit into beginTransaction-runQueries-commitTransaction pattern, and it's not atomic anyway. It also does not happen to be supported asynchronously in the client library (although it is in Connection API, but more as a wrap-this-in-a-thread workaround). 

cc/ @olavloite

Fixes #249 .

### DIFF
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
@@ -21,21 +21,15 @@ import com.google.api.core.ApiFutureCallback;
 import com.google.api.core.ApiFutures;
 import com.google.cloud.spanner.AsyncResultSet;
 import com.google.cloud.spanner.AsyncResultSet.CallbackResponse;
-import com.google.cloud.spanner.AsyncTransactionManager;
-import com.google.cloud.spanner.AsyncTransactionManager.AsyncTransactionStep;
-import com.google.cloud.spanner.AsyncTransactionManager.TransactionContextFuture;
 import com.google.cloud.spanner.DatabaseAdminClient;
 import com.google.cloud.spanner.DatabaseClient;
 import com.google.cloud.spanner.DatabaseId;
 import com.google.cloud.spanner.ReadContext;
-import com.google.cloud.spanner.ReadOnlyTransaction;
 import com.google.cloud.spanner.Spanner;
 import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.TimestampBound;
 import com.google.cloud.spanner.TransactionContext;
 import com.google.cloud.spanner.r2dbc.SpannerConnectionConfiguration;
-import com.google.spanner.v1.TransactionOptions;
-import com.google.spanner.v1.TransactionOptions.ReadWrite;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
@@ -67,15 +61,7 @@ class DatabaseClientReactiveAdapter {
 
   private final ExecutorService executorService;
 
-  // transaction context
-
-  private AsyncTransactionManager transactionManager;
-
-  private TransactionContextFuture txnContextFuture;
-
-  private ReadOnlyTransaction readOnlyTransaction;
-
-  private AsyncTransactionStep<?, ? extends Object> lastStep;
+  private final DatabaseClientTransactionManager txnManager;
 
   /**
    * Instantiates the adapter with given client library {@code DatabaseClient} and executor.
@@ -93,12 +79,7 @@ class DatabaseClientReactiveAdapter {
     this.dbAdminClient = spannerClient.getDatabaseAdminClient();
     this.executorService = executorService;
     this.config = config;
-  }
-
-  private boolean isInReadWriteTransaction() {
-    // TODO: do we need to close a read-only transaction?
-
-    return this.txnContextFuture != null;
+    this.txnManager = new DatabaseClientTransactionManager(this.dbClient, this.executorService);
   }
 
   /**
@@ -107,23 +88,22 @@ class DatabaseClientReactiveAdapter {
    * @return reactive pipeline for starting a transaction
    */
   public Mono<Void> beginTransaction() {
-    return convertFutureToMono(() -> {
-      LOGGER.debug("begin readwrite transaction");
-      this.transactionManager = this.dbClient.transactionManagerAsync();
-      this.txnContextFuture = this.transactionManager.beginAsync();
-      return this.txnContextFuture;
-    }).then();
+    return convertFutureToMono(() -> this.txnManager.beginTransaction()).then();
   }
 
+  /**
+   * Allows starting a Cloud Spanner readonly transaction with custom staleness.
+   *
+   * @return reactive pipeline for starting a transaction
+   */
   public Mono<Void> beginReadonlyTransaction(TimestampBound timestampBound) {
 
-    // TODO: what if another transaction is in progress?
-    return Mono.defer(
-        () -> {
-            System.out.println("********* READONLY TRANSACTION detected");
-            this.readOnlyTransaction = this.dbClient.readOnlyTransaction(timestampBound);
-            return Mono.empty();
-        });
+    return Mono.defer(() -> {
+      this.txnManager.beginReadonlyTransaction(timestampBound);
+      // TODO: it would be good for client library to signal when a session is ready for this
+      // transaction, so a meaningful mono is returned.
+      return Mono.empty();
+    });
   }
 
   /**
@@ -133,38 +113,9 @@ class DatabaseClientReactiveAdapter {
    */
   public Publisher<Void> commitTransaction() {
 
-    return Mono.defer(() -> {
-      if (isInReadWriteTransaction()) {
-        return convertFutureToMono(() -> {
-          LOGGER.debug("commit transaction");
-
-          if (this.lastStep == null) {
-            // TODO: replace by a better non-retryable;
-            //  consider not throwing at all and no-oping with warning.
-            throw new RuntimeException("Nothing was executed in this transaction");
-          }
-          return this.lastStep.commitAsync();
-          });
-      } else if (this.readOnlyTransaction != null) {
-        // TODO: does transaction need to be closed to release it back to the pool?
-        this.readOnlyTransaction = null;
-        return Mono.empty();
-      } else {
-        // TODO: log warning that commit was called without begin? or check sPI for expected behavior
-        return Mono.empty();
-      }
-
-    }).doOnTerminate(this::clearTransactionManager).then();
-  }
-
-  private void clearTransactionManager() {
-    LOGGER.debug("close transaction manager");
-    this.txnContextFuture = null;
-    this.lastStep = null;
-    if (this.transactionManager != null) {
-      this.transactionManager.close();
-      this.transactionManager = null;
-    }
+    return convertFutureToMono(() -> this.txnManager.commitTransaction())
+        .doOnTerminate(this.txnManager::clearTransactionManager)
+        .then();
   }
 
   /**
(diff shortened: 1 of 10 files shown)


cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
```diff
@@ -97,12 +88,22 @@ private boolean isInTransaction() {
...
-    }).then();
+    return convertFutureToMono(() -> this.txnManager.beginTransaction()).then();
+  }
+
+  /**
+   * Allows starting a Cloud Spanner readonly transaction with custom staleness.
+   *
+   * @return reactive pipeline for starting a transaction
+   */
+  public Mono<Void> beginReadonlyTransaction(TimestampBound timestampBound) {
```
meltsufin, line 99: How does one specify the preference to do a strong read? I.e. no timestamp.

meltsufin (commented):
Overall, looks good. Just one questions about allowing read-only without specifying time bound.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientTransactionManager.java
```diff
@@ -0,0 +1,168 @@
...
+   *     The function accepts ReadContext for SELECT queries, and TransactionContext for DML.
+   * @param <T> Type of object wrapped by the {@link ApiFuture} returned by the operation
+   * @return {@link ApiFuture} result of the provided operation
+   */
+  <T> ApiFuture<T> runInTransaction(Function<? super TransactionContext, ApiFuture<T>> operation) {
+
+    // The first statement in a transaction has no input, hence Void input type.
+    // The subsequent statements take the previous statements' return (affected row count)
+    // as input.
+    AsyncTransactionStep<? extends Object, T> updateStatementFuture =
```
dmitry-s, line 153: Maybe remove this intermediate variable?

dmitry-s (approved):
LGTM, one small suggestion

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientTransactionManager.java
elefeint: Can't. The temporary function reference has `T` return type, but the field is `Object`. The more specific type has to get returned.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
elefeint: Two ways: 
1) going without a transaction (default single-use context is for strong reads)
2) `TimestampBound.strong()`

meltsufin: for (1), would that be read-write or read-only transaction?

elefeint: We split by type of statement, so without an explicit transaction, a SELECT statement would run through `DatabaseClientReactiveAdapter.runSelectStatement()`, which would pick a single-use read transaction with strong consistency.

If an explicit read-write transaction is started, then the select statements will run inside of the same transaction (also strong, but as of transaction start).

meltsufin: I'm still not clear on how do I explicitly start a read-only transaction without a timestamp bound, and for multiple statements? Does this question even make sense?

elefeint: It does make sense -- you'd use `.beginReadonlyTransaction(TimestampBound.strong())`

https://github.com/googleapis/java-spanner/blob/fb7d935b9af219ee663564a134f87c09b342f64d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TimestampBound.java#L139

meltsufin: Got it. So, you don't feel it makes sense to provide an overloaded `.beginReadonlyTransaction()` which would be the same as `.beginReadonlyTransaction(TimestampBound.strong())`.

meltsufin (approved):

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientReactiveAdapter.java
elefeint: That may be a convenient shortcut, since it's not super intuitive that you have to *use* TimestampBound to avoid a timestamp bound.
Happy to add it. Thursday :) I am a sufficient amount of wine into the election results that I will probably trip over a semicolon on my way to a commit.

olavloite: +1 for adding the convenience method for `.beginReadonlyTransaction()` for a strong read-only transaction. The client library also provides the same kind of convenience method.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientTransactionManager.java
```diff
@@ -0,0 +1,168 @@
...
+  boolean isInReadWriteTransaction() {
+    return this.txnContextFuture != null;
+  }
+
+  /**
+   * Starts a Cloud Spanner Read/Write transaction.
+   *
+   * @return chainable {@link TransactionContextFuture} for the current transaction.
+   */
+  TransactionContextFuture beginTransaction() {
```
dzou, line 71: Do you need to worry about if someone calls `beginTransaction()` and then call `beginReadonlyTransaction()` after? (or shall we assume this is undefined behavior?)

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/it/ClientLibraryBasedIT.java
```diff
@@ -610,6 +614,43 @@ public void insertMultipleBoundParameterSetsInTransaction() {
...
+        .expectNext(1L)
+        .as("strong read returns the inserted row without a transaction")
+        .expectNext(0L)
+        .as("stale read returns nothing")
+        .expectNext(1L)
+        .as("strong read returns the inserted row after stale-read transaction terminates")
+        .verifyComplete();
+  }
+
+  private String concatRow(Row row) {
```
dzou, line 650: This method can be removed; doesn't appear to be used anywhere.

dzou (dismissed):

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientTransactionManager.java
elefeint: Good question, a little bit related to your autocommit issue. R2DBC spec says that if transaction autocommit state is changed, the transaction gets committed. It also says that if transaction isolation changes (which is the concept most similar to changing to/from readonly), it's up to the implementation.
I'd probably follow the autocommit principle here -- if a new transaction is attempted while the old one is in progress, commit everything up to this point.

olavloite: The JDBC spec says basically the same thing (i.e. commit current transaction when changing autocommit, up to the implementation when changing read-only). The Cloud Spanner JDBC driver uses this freedom to act as strictly as possible:
* Changing autocommit during a transaction is allowed, as per spec.
* Changing read-only during a transaction causes an error.

dzou: I see, makes sense. I can implement this feature for autocommit.

elefeint: Let me test it first; from what @olavloite said it looks like we may be able to rely on the underlying client library behavior.

elefeint: @olavloite Calling `databaseClient.transactionManagerAsync().beginAsync()` followed by `this.dbClient.readOnlyTransaction()` does not fail -- @meltsufin theorized that it's because the client library picks up a different session for the second command.
For the purpose of R2DBC driver, I introduced an explicit check to prevent multiple simultaneous transaction attempts.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientTransactionManager.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+  }
+
+  /**
+   * Commits the current transaction (if read/write) or closes it (if read-only).
+   *
+   * @return chainable {@link ApiFuture} for commit status.
+   */
+  ApiFuture<? extends Object> commitTransaction() {
+    if (isInReadWriteTransaction()) {
```
dzou, line 121: Should `clearTransactionManager()` be called in here now instead of chained to `commit()` in the adapter? Similar to what is done with `closeReadOnlyTransaction()`

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientTransactionManager.java
olavloite: For the R2DBC driver that is something that you probably want to prohibit, but in the client library it is legal. A `DatabaseClient` is not treated like a connection or other 'exclusive' resource, and an application may request multiple parallel transactions from it, and these may also be of different types. So the following simplified example is perfectly fine, as the `DatabaseClient` will as @meltsufin theorized use different sessions for each transaction:

```java
    ReadOnlyTransaction readOnlyTx = dbClient.readOnlyTransaction();
    TransactionManager mgr = dbClient.transactionManager();
    TransactionContext context = mgr.begin();
    try (ResultSet rs = readOnlyTx.executeQuery(Statement.of("SELECT * FROM FOO"))) {
      long val = rs.getLong("COL1");
      context.executeUpdate(
          Statement.newBuilder("UPDATE BAR SET COL2=@val").bind("val").to(val).build());
    }
    mgr.commit();
    readOnlyTx.close();
```

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientTransactionManager.java
elefeint: So, that would be cleaner, but because `DatabaseClientTransactionManager` returns futures, we don't know whether the operation succeeded or failed in those commit/rollback methods. I could start adding Future callbacks to it, but it seems more error-prone than attaching something to the end of a reactive pipeline.

On the other hand, right now `DatabaseClientReactiveAdapter` uses `doOnTerminate()`, which will clear the transaction state even upon failure. I filed #288 to look into that.

dzou: Sounds good.

dzou (approved):

PR #291, opened 2020-11-17
support case-insensitive column names

### LINKED ISSUE #271: Revisit case-sensitive column implementation
Cloud Spanner column names are case-sensitive. We've historically been overriding the `columnMetadata()` and `duplicateColumnNames()` TCK test methods to reflect that case sensitivity is expected.

However, this implementation technically violates the R2DBC `Row` and `RowMetadata` specs that say:

> Column names are case insensitive.  When a get method contains several columns with same name, then the value of the first 
 matching column will be returned
>     * @return the {@link ColumnMetadata} for one column in this row

Consider whether for V2, we should revert to R2DBC spec definition of case-insensitive column names.

dmitry-s:
fixes #271

### DIFF
--- a/cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryRowMetadataTest.java
+++ b/cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryRowMetadataTest.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2020-2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryRowMetadataTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    ColumnMetadata column1 = metadata.getColumnMetadata("column_1");
+    assertThat(column1.getNativeTypeMetadata())
+        .isEqualTo(Type.string());
+
+    ColumnMetadata column0 = metadata.getColumnMetadata(0);
+    assertThat(column0.getName()).isEqualTo("column_0");
+    assertThat(column0.getNativeTypeMetadata()).isEqualTo(Type.int64());
+
+    // When a get method contains several columns with same name, then the value of the first
```
elefeint, line 80: I'd split this into a separate test.

elefeint (dismissed):
Good tests! One optional thing.

elefeint (approved):

PR #293, opened 2020-11-23
Add config options to mirror JDBC driver

### LINKED ISSUE #248: Investigate passing URL options to client library
Investigate where the options/URL need to be passed to spanner for consistent behavior with JDBC driver. Make sure the same options would be supported in JDBC vs R2DBC.

  If necessary, reassemble R2DBC options into a URL to pass to client library (R2DBC parses URL and does not retain the original string in the option list).

elefeint:
Getting this absolutely cursed (by Murphy's law, life, and random rabbitholes) PR out.

I still need to add `autocommit` and `readonly`.

Some properties are duplicated with the client library Connection API; those can be addressed incrementally. I'll also look to refactor a bit of Connection API's property initialization to avoid duplicate work of parsing with R2DBC SPI and then re-parsing in Connection API. But non of these is blocking for V2 release, seeing as Connection API is something of a nice-to-have.

Fixes #248.

### DIFF
--- a/cloud-spanner-r2dbc/README_v2.adoc
+++ b/cloud-spanner-r2dbc/README_v2.adoc
@@ -15,6 +15,7 @@ ConnectionFactories.get(
 ## Authentication
 
 The driver allows the following options for authentication:
+
 * a `String` property `credentials` containing the local file location of the JSON credentials file.
 * a `Credentials` object provided as `google_credentials`. This will only work with programmatically constructed `ConnectionFactoryOptions`.
 * a `String` OAuth token provided as `oauthToken`.
--- a/cloud-spanner-r2dbc/README_v2.adoc
+++ b/cloud-spanner-r2dbc/README_v2.adoc
@@ -22,6 +22,53 @@ The driver allows the following options for authentication:
 If no authentication options are provided, Application Default Credentials will be automatically determined.
 The only exception is when the connection is in plain-text, indicating the use of Cloud Spanner emulator, in which case no credentials will be used.
 
+## Supported connection options
+
+All connection options of primitive and String type can be passed through the connection URL in the `?key1=value1&key2=value2` format.
+Object-typed options can only be passed in programmatically.
+
+|===
+|Property name |Type |Allowed in URL connection |Default |Comments
+
+|`credentials`
+|String
+|Yes
+|null
+|The location of the credentials file to use for this connection
+
+|`oauthToken`
+|String
+|Yes
+|null
+|A valid pre-existing OAuth token to use for authentication
+
+|`google_credentials`
+|com.google.auth.oauth2.OAuth2Credentials
+|No
+|null
+|A pre-authenticated authentication object that can only be supplied with programmatic connection options
+
+|`thread_pool_size`
+|int
+|Yes
+|Equal to the number of available processors
+|Determines the size of thread pool used to process `Future` callbacks
+
+|`usePlainText`
+|boolean
+|Yes
+|false
+|Turns off SSL and credentials use (only valid when using Cloud Spanner emulator)
+
+|optimizerVersion
+|String
+|Yes
+|null
+= ``|``Determines version of Cloud Spanner https://cloud.google.com/spanner/docs/query-optimizer/query-optimizer-versions[optimizer] to use in queries
+
+|===
+
+
 ## Transactions
 
 ### Read-Write Transactions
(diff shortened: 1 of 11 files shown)


cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
```diff
@@ -66,14 +73,34 @@
...
+  public static final Option<String> OPTIMIZER_VERSION =
+      Option.valueOf("optimizerVersion");
   /**
-   * Option specifying the location of the GCP credentials file.
+   * Option specifying the already-instantiated credentials object.
    */
   public static final Option<GoogleCredentials> GOOGLE_CREDENTIALS =
       Option.valueOf("google_credentials");
 
+  /** Option specifying the location of the GCP credentials file. Same as GOOGLE_CREDENTIALS,
```
dzou, line 85: Fix javadoc formatting here.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
```diff
@@ -66,14 +73,34 @@
...
+   * but consistent with the JDBC driver option.
+   */
+  public static final Option<String> CREDENTIALS = Option.valueOf(CREDENTIALS_PROPERTY_NAME);
```
dzou, line 88: Do the credentials options need to have the same name as what's defined in the client libraries? Feel like maybe just having a regular String might be simpler than having to refer back to other code to find out what the value is.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/AbstractSpannerClientLibraryStatement.java
```diff
@@ -55,8 +55,17 @@
   public AbstractSpannerClientLibraryStatement(
       DatabaseClientReactiveAdapter clientLibraryAdapter, String query) {
     this.clientLibraryAdapter = clientLibraryAdapter;
-    this.currentStatementBuilder = com.google.cloud.spanner.Statement.newBuilder(query);
     this.query = query;
+    this.currentStatementBuilder = createStatementBuilder();
+  }
+
+  private com.google.cloud.spanner.Statement.Builder createStatementBuilder() {
```
dzou, line 62: Would prefer this to be a `static` method that accepted 2 parameters - this way you will never make the mistake of accidentally initializing `currentStatementBuilder` before the other two values are set.

dzou (approved):

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/CredentialsHelper.java
```diff
@@ -0,0 +1,58 @@
...
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Credential instantiation service that can interact with the filesystem.
+ */
+class CredentialsHelper {
+
+  GoogleCredentials getOauthCredentials(String oauthToken) {
```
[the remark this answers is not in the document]
elefeint: I split this off into a helper class to make it mockable for unit tests -- don't want to interact with the filesystem in those.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProvider.java
elefeint: They should have the same name, but we don't really need to refer back to Connection API. I'll hardcode them. In the even the values change in Connection API, we'd want to manually change docs anyway.

Speaking of docs, let me go and document the allowed properties.

cloud-spanner-r2dbc/README_v2.adoc
```diff
@@ -12,6 +12,63 @@ ConnectionFactories.get(
               .build());
 ```
 
+## Authentication
+
+The driver allows the following options for authentication:
```
dzou, line 17:
```suggestion
The driver allows the following options for authentication:

```

dzou (dismissed):

meltsufin (dismissed):
Looks good!

dzou (approved):

PR #298, opened 2020-11-24
Restrict visibility of public V2 implementation classes

elefeint:
This PR restricts visibility of most v2 implementation classes because in the future, `.v2` package contents will move one level up, so nobody should be relying on the `.v2` classes directly. I moved type conversion classes up a level into `.v2` package to be able to reduce their visibility, too.

The only public class left is `SpannerClientLibraryConnectionFactory`, which needs to be visible to `SpannerConnectionFactoryProvider`. 

I moved all V2 exceptions up a level, since those have a legit reason to being referenced by end users.

Since we need to give users access to custom `SpannerClientLibraryConnection` methods for creating readonly transactions, `com.google.cloud.spanner.r2dbc.api.SpannerConnection` is a new interface allowing access to those methods without exposing `SpannerClientLibraryConnection` itself.

There are also a couple of integration tests that were referring to internal implementation details, so I moved one to become a unit test, and refactored the other to rely on SPI methods instead.

### DIFF
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/api/SpannerConnection.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/api/SpannerConnection.java
@@ -19,6 +19,13 @@ package com.google.cloud.spanner.r2dbc.api;
 import com.google.cloud.spanner.TimestampBound;
 import reactor.core.publisher.Mono;
 
+/**
+ * Interface representing custom Cloud Spanner Connection transaction options that do not fit the
+ * current R2DBC SPI model.
+ *
+ * <p>Cast a Cloud Spanner-provided `Connection` object to this interface to take advantage of the
+ * readonly transaction types: read-only strongly consistent and read-only stale.
+ */
 public interface SpannerConnection {
   /**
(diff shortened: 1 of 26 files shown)


cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/api/SpannerConnection.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner.r2dbc.api;
+
+import com.google.cloud.spanner.TimestampBound;
+import reactor.core.publisher.Mono;
+
+public interface SpannerConnection {
```
meltsufin, line 22: Do we need a class-level javadoc comment?

meltsufin (approved):

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/api/SpannerConnection.java
elefeint: Yes; I missed it. Not sure how checkstyle let me get away with that.

meltsufin (approved):

PR #329, opened 2021-03-16
Add GHA config file for integration tests

elefeint:


dzou (approved):
Nice! Do you want to remove the Travis CI yaml + encrypted key in this PR too?

.github/workflows/integrationTests.yaml
```diff
@@ -0,0 +1,101 @@
...
+  schedule:
+    - cron: '00 8 * * *' # 08:00 UTC every day
+
+
+jobs:
+  integrationTests:
+    if: github.repository == 'GoogleCloudPlatform/cloud-spanner-r2dbc'
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: true
```
ttomsu, line 17: You can remove the `strategy` block if not doing any matrix builds

ttomsu (approved):

ttomsu:
You may need to just rerun the integration tests.

elefeint:
It was a real problem that travis integration tests were masking! I had made this really questionable choice last year to avoid recreating tables by default. The problem is, that depending on the order in which the tests run, they need different column definitions. In the long run, these tests should either standardize the table schema required or use different tables. But for now, defaulting to recreating tables fixes the problem.

PR #402, opened 2021-07-09
Switch default implementation to v2

elefeint:
* Officially switch the default implementation to client-library-based
* Merge docs, leave a short-ish blurb for v1 specifics on the bottom
* Update leftover "spanner:" protocol usages to "cloudspanner:" except in tests explicitly verifying that old protocol remains supported for now.
* Remove unused `threadPoolSize` settings from V2.

Biggest change is in the docs. We'll do more passes over the docs, but it would be good to validate that they are coherent.


README.md
```diff
@@ -1,10 +1,9 @@
 # Cloud Spanner R2DBC Driver
 
-[![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
-
-An implementation of the [R2DBC](https://r2dbc.io/) driver for [Cloud Spanner](https://cloud.google.com/spanner/) is being developed in this repository.
-
-A [Spring Data R2DBC dialect for Cloud Spanner](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/tree/main/cloud-spanner-spring-data-r2dbc) dialect is available, as well.
+This project contains:
+* An implementation of Java Reactive Relational Database Connectivity SPI [R2DBC](https://r2dbc.io/) for [Cloud Spanner](https://cloud.google.com/spanner/) based on the Cloud Spanner [client library](https://github.com/googleapis/java-spanner).
+* A [Spring Data R2DBC dialect for Cloud Spanner](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/tree/main/cloud-spanner-spring-data-r2dbc).
```
meltsufin, line 5: It's a separate dependency, right?

How do we guide users to Spring Data R2DBC more? There doesn't seem to be much discussion on it in this doc.

README.md
```diff
@@ -211,55 +203,48 @@ Spanner offers [three transaction types](https://cloud.google.com/spanner/docs/t
...
+                              conn.commitTransaction(),
+                        )
+    ```
+    NOTE: Readonly transactions must be closed by calling `commit()` before starting a new read-write or a read-only transaction.
+
 
 - Partitioned DML: A transaction designed for bulk updates and deletes with certain restrictions.
     See the [Partitioned DML documentation](https://cloud.google.com/spanner/docs/dml-partitioned)
     for more information.
+    This driver does not support Partitioned DML transactions at the time.
```
meltsufin, line 241: Is there an issue for this?

README.md
```diff
@@ -314,43 +295,11 @@ It will produce a `Publisher` (implemented by a `Flux`) containing two `SpannerR
...
-requests from Spanner in it's first request to store in the buffer. This number is an upper
-bound that affects how it is replenished; when the client has consumed 25% of the PartialResultSets,
-it will request 25% upstream from Spanner. See the [prefetch documentation](https://projectreactor.io/docs/core/release/reference/#_operators_that_change_the_demand_from_downstream)
-for more info.
+## Backpressure
 
-The default buffer size of `PartialResultSets` is 1 and this may be increased by setting
-`partial_result_set_fetch_size`.
+Backpressure on SQL SELECT queries is supported out of the box.
+Take care to ultimately always exhaust or cancel the query result `Publisher`, since not doing so may lead to objects not being deallocated properly.
```
meltsufin, line 302:
```suggestion
Take care to always ultimately exhaust or cancel the query result `Publisher`, since not doing so may lead to objects not being deallocated properly.
```

README.md
```diff
@@ -400,47 +348,76 @@ Flux.from(connection.createBatch()
...
-```xml
-<dependency>
-  <groupId>io.r2dbc</groupId>
-  <artifactId>r2dbc-pool</artifactId>
-  <version>1.0.0.BUILD-SNAPSHOT</version>
-</dependency>
-```
-**Example:**
+
+## gRPC-based driver (v1)
```
meltsufin, line 356: What's the rationale for still including v1? When is it going to be removed?

README.md
```diff
@@ -400,47 +348,76 @@ Flux.from(connection.createBatch()
...
+| `database`  | Your Spanner Database name | True (if `url` not provided) | |
+| `url`       | A Cloud Spanner R2DBC URL specifying your Spanner database. An alternative to specifying `project`, `instance`, and `database` separately. | False |
+| `google_credentials` | Optional [Google credentials](https://cloud.google.com/docs/authentication/production) override to specify for your Google Cloud account. | False | If not provided, credentials will be [inferred from your runtime environment](https://cloud.google.com/docs/authentication/production#finding_credentials_automatically).
+| `partial_result_set_fetch_size` | Number of intermediate result sets that are buffered in transit for a read query. | False | 1 |
+| `ddl_operation_timeout` | Duration in seconds to wait for a DDL operation to complete before timing out | False | 600 seconds |
+| `ddl_operation_poll_interval` | Duration in seconds to wait between each polling request for the completion of a DDL operation | False | 5 seconds |
+
+### DDL Statements in v1
+
+DDL statements in Spanner receive special treatment by Cloud Spanner. Creating and
```
meltsufin, line 389: Are you using the one-line-per-sentence style? I see in other parts of the doc.

meltsufin (commented):
The v1 and v2 discussion in the same readme seems to be a bit confusing to me. What's the plan to ultimately remove the v1 baggage?

README.md
```diff
@@ -37,15 +36,84 @@ The easiest way to start using the driver is to add the driver dependency throug
...
 ```
 
+### Usage
+
+After setting up the dependency and [authentication](#authentication), you can begin directly using the driver.
+
+The rest of this documentation will show examples of directly using the driver.
+In a real application, you should use one of R2DBC's user-friendly [client APIs](https://r2dbc.io/clients/) instead.
+
+The entry point to using the R2DBC driver is to first configure the R2DBC connection factory.
```
dzou, line 50:
```suggestion
The entrypoint to using the R2DBC driver is to first configure the R2DBC connection factory.
```

Is entrypoint one word?

README.md
```diff
@@ -74,89 +141,24 @@ special type of Google Account intended to represent a non-human user that needs
...
+|-------------------------|-------|--------------------------|--------|--------|
+|`credentials`            |String |Yes                       |null    |The location of the credentials file to use for this connection
+|`oauthToken`             |String |Yes                       |null    |A valid pre-existing OAuth token to use for authentication
+|`google_credentials`     |com.google.auth.oauth2.OAuth2Credentials|No|null|A pre-authenticated authentication object that can only be supplied with programmatic connection options
+|`usePlainText`           |boolean|Yes                       |false   |Turns off SSL and credentials use (only valid when using Cloud Spanner emulator)
+|`optimizerVersion`       |String |Yes                       |null    |Determines version of Cloud Spanner https://cloud.google.com/spanner/docs/query-optimizer/query-optimizer-versions[optimizer] to use in queries
+|`autocommit`             |boolean|Yes                       |true    |Whether new connections are created in autocommit mode
+|`readonly`               |boolean|Yes                       |false   | Whether new connections start with a read-only transaction
```
dzou, line 161:
```suggestion
|`readonly`               |boolean|Yes                       |false   |Whether new connections start with a read-only transaction
```

dzou (approved):

README.md
elefeint: Haphazardly. I'll update to that style throughout.

README.md
elefeint: We can just remove it as part of this change, I suppose. There is no instance in which we'd want customers using v1.

README.md
elefeint: Right; I want to write a blurb about this also, as a follow-up PR. We have the sample, but it would be good to show in the docs.

The reason there was not any discussion of Spring Data before is that none of it is specific to Cloud Spanner driver.

README.md
elefeint: I've created #406. Also forwarded a discussion from last year to the team.

README.md
elefeint: I'll just replace it by "To start".

README.md
elefeint: #407 to track.

PR #414, opened 2021-07-15
Implemented batching for DML

### LINKED ISSUE #409: Support R2DBC batch in v2
https://r2dbc.io/spec/0.8.0.M8/spec/html/#batches.interface

elefeint:
This PR brings batch support to the same level that we had in v1 -- supporting batches of DML only.

R2DBC batches were meant for interspersed DML and SELECT. However, Cloud Spanner's `executeBatchDml` endpoint [only allows DML statements in the batch](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.ExecuteBatchDmlRequest).

As part of implementation, the mapping from a `Mono<long[]>` to an R2DBC `Result` moved from the statement implementation to the `DatabaseClientReactiveAdapter`. It's an adapter from futures to R2DBC primitives... let it adapt fully instead of going halfway to an intermediate `Mono<long[]>`

Fixes #409

### DIFF
--- a/cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerClientLibraryTestKit.java
+++ b/cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerClientLibraryTestKit.java
@@ -280,15 +280,12 @@ public class SpannerClientLibraryTestKit implements TestKit<String> {
   @Test
   public void batch() {
     // Only DML statements are accepted by ExecuteBatchDml endpoint
-
     Flux.usingWhen(getConnectionFactory().create(),
         connection -> Flux.from(connection
-
             .createBatch()
             .add(expand(TestStatement.INSERT_VALUE200))
-            .add(expand(TestStatement.INSERT_VALUE100)) // updatd from SELECT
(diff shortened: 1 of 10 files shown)


cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerClientLibraryTestKit.java
```diff
@@ -277,12 +277,23 @@ public void clobSelect() {
...
   @Test
   public void batch() {
-    /*
-    batch DML support is not supported currently.
-     */
+    // Only DML statements are accepted by ExecuteBatchDml endpoint
+
+    Flux.usingWhen(getConnectionFactory().create(),
+        connection -> Flux.from(connection
+
```
dzou, line 286: Are these newlines intended?

dzou (approved):

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/it/SpannerClientLibraryTestKit.java
elefeint: Not really; I copy/pasted and did not think about it.

PR #416, opened 2021-07-16
Add Java type lookup

### LINKED ISSUE #206: Implement SpannerColumnMetadata.getJavaType()
SPI recommends implementing this method; we've missed it.

elefeint:
Fixes #206


meltsufin (approved):
Crazy how much code got eliminated!
Any way to fix that code smell regarding generics?
Otherwise, LGTM.

elefeint:
`<?>` would be the only reasonable type, since there is no commonality otherwise
To be fair, I just missed deleting this code in the blast-5K-LOC PR.

PR #467, opened 2021-12-06
Use the new authentication way

elefeint:
Gcloud separated the auth step into a different action. Workload federation is recommended, but _(insert long story, available on demand)_ I am sticking with the path of least resistance and using the same JSON credential.


chanseokoh (approved):

.github/workflows/integrationTests.yaml
```diff
@@ -28,12 +28,16 @@ jobs:
         with:
           path: ~/.m2/repository
           key: ${{ runner.os }}-maven-unified-${{ steps.date.outputs.date }}
+      - name: Set Up Authentication
+        id: auth
```
chanseokoh, line 32: nit: probably `id` is unnecessary?

Issue #519, opened 2022-03-30
Missing binders for java.time.LocalDateTime and java.time.LocalDate types


magiccrafter:
Writing to DATE and TIMESTAMP columns fails using the R2dbcEntityTemplate abstraction for fields of LocalDate and/or LocalDateTime java.time types.

Stacktrace:
```
Can't find a binder for type: class java.time.LocalDateTime; nested exception is com.google.cloud.spanner.r2dbc.BindingFailureException: Can't find a binder for type: class java.time.LocalDateTime
	at org.springframework.r2dbc.connection.ConnectionFactoryUtils.convertR2dbcException(ConnectionFactoryUtils.java:238)
	at org.springframework.r2dbc.core.DefaultDatabaseClient.lambda$inConnectionMany$6(DefaultDatabaseClient.java:142)
	at reactor.core.publisher.FluxUsingWhen.deriveFluxFromResource(FluxUsingWhen.java:120)
	at reactor.core.publisher.FluxUsingWhen.access$000(FluxUsingWhen.java:54)
	at reactor.core.publisher.FluxUsingWhen$ResourceSubscriber.onNext(FluxUsingWhen.java:193)
	at reactor.core.publisher.FluxMap$MapSubscriber.onNext(FluxMap.java:120)
	at reactor.core.publisher.FluxOnErrorResume$ResumeSubscriber.onNext(FluxOnErrorResume.java:79)
	at reactor.core.publisher.FluxOnErrorResume$ResumeSubscriber.onNext(FluxOnErrorResume.java:79)
	at reactor.core.publisher.FluxRetry$RetrySubscriber.onNext(FluxRetry.java:87)
	at reactor.core.publisher.Operators$MonoSubscriber.complete(Operators.java:1816)
	at reactor.core.publisher.MonoFlatMap$FlatMapInner.onNext(MonoFlatMap.java:249)
	at io.r2dbc.pool.MonoDiscardOnCancel$MonoDiscardOnCancelSubscriber.onNext(MonoDiscardOnCancel.java:92)
...
```

The necessary data converters are missing for both:
- java.time.LocalDate
- java.time.LocalDateTime

This issue requires the following to be implemented first: https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/issues/517

magiccrafter:
includes the changes for #517

Issue #517, opened 2022-03-29
Add google java types to the spring data's SimpleType collection in the SpannerR2dbcDialect


magiccrafter:
google's Spanner java types have to be added to the [SpannerR2dbcDialect](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/blob/02e1d0b9f3566f5e1b53887f39b2c155eb61320f/cloud-spanner-spring-data-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/springdata/SpannerR2dbcDialect.java#L99)'s `getSimpleTypes()` collection in order to create custom [@WritingConverter ](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/convert/WritingConverter.html) converters to map other types to them:
- `com.google.cloud.Timestamp`
- `com.google.cloud.ByteArray`
- `com.google.cloud.Date`

Based on the Spring FW's [BasicRelationalConverter#writeValue](https://github.com/spring-projects/spring-data-relational/blob/f2dc64ec52e361acdd5022032fb9dde27b9ca7bd/spring-data-relational/src/main/java/org/springframework/data/relational/core/conversion/BasicRelationalConverter.java#L157) converter core logic the custom spring-data client libraries should have its java types added to the SimpleType list in their implementation of the `R2dbcDialect`. 

Below is the snippet from `BasicRelationalConverter` where the key logic resides:
```java
@Override
@Nullable
public Object writeValue(@Nullable Object value, TypeInformation<?> type) {
    if (value == null) {
        return null;
    }

    if (getConversions().isSimpleType(value.getClass())) {
        if (ClassTypeInformation.OBJECT != type) {
            if (conversionService.canConvert(value.getClass(), type.getType())) {
                value = conversionService.convert(value, type.getType());
            }
        }
        return getPotentiallyConvertedSimpleWrite(value);
    }

    RelationalPersistentEntity<?> persistentEntity = context.getPersistentEntity(value.getClass());
    if (persistentEntity != null) {
        Object id = persistentEntity.getIdentifierAccessor(value).getIdentifier();
            return writeValue(id, type);
    }

    return conversionService.convert(value, type.getType());
}
```

magiccrafter:
@elefeint you can close that one, thanks

PR #520, opened 2022-03-30
Missing binders for java.time.LocalDateTime and java.time.LocalDate t…

magiccrafter:
…ypes (#519)

### DIFF
--- a/cloud-spanner-spring-data-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/springdata/it/SpannerR2dbcDialectIntegrationTest.java
+++ b/cloud-spanner-spring-data-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/springdata/it/SpannerR2dbcDialectIntegrationTest.java
@@ -119,7 +119,7 @@ class SpannerR2dbcDialectIntegrationTest {
                 president.getName().equals("Bill Clinton")
                     && president.getStartYear() == 1992
                     && president.getStartDate() == LocalDate.parse("1992-01-01")
-                    && president.getCreatedAt() == LocalDateTime.parse("1992-01-01T00:00:00"))
+                    && president.getStartDate().equals(LocalDate.parse("1992-01-01")
         .verifyComplete();
   }
 
--- a/cloud-spanner-spring-data-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/springdata/it/SpannerR2dbcDialectIntegrationTest.java
+++ b/cloud-spanner-spring-data-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/springdata/it/SpannerR2dbcDialectIntegrationTest.java
@@ -27,12 +27,12 @@ import io.r2dbc.spi.ConnectionFactories;
 import io.r2dbc.spi.ConnectionFactory;
 import io.r2dbc.spi.ConnectionFactoryOptions;
 import io.r2dbc.spi.Option;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.domain.Sort.Direction;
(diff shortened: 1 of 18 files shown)


zhumin8:
Thanks for catching this and contributing!
Looks like there are some checkstyle errors blocking the tests. You might find the google-java-format tool mentioned [here](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/blob/main/CONTRIBUTING.adoc#style) helpful.

magiccrafter:
It should be ok now 🤞. I've added the requested integration tests for #517 here.

magiccrafter:
@elefeint Is there anything else I can help with for this MR?

off-topic: Out of curiosity, when are you planning to release next? I couldn't find any relevant Milestones, and I've seen that the main branch has quite a lot of changes since the last release (1.1.0).

magiccrafter:
@elefeint Is there anything else I can help with for this MR?

off-topic: Out of curiosity, when are you planning to release next? I couldn't find any relevant Milestones, and I've seen that the main branch has quite a lot of changes since the last release (1.1.0).

elefeint:
The codebase is halfway through switchover to the latest version of the spec, 0.9, and the upcoming 1.0 release of the R2DBC SPI. It's blocked on me right now. You are right that it would be good to do a release soon, but unfortunately I can't promise dates.

cloud-spanner-spring-data-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/springdata/it/SpannerR2dbcDialectIntegrationTest.java
```diff
@@ -105,41 +100,38 @@ public void initializeTestEnvironment() {
...
-            .rowsUpdated()
-            .block();
+        .sql("DELETE FROM PRESIDENT where NAME is not null")
+        .fetch()
+        .rowsUpdated()
+        .block();
   }
 
   @Test
   void testReadWrite() {
```
zhumin8, line 110: Unfortunately we don't run integration tests on external contributions, you will need to setup your own project to test this. 
I tried to run this with the comparison and constructor fixed, and find it fails with error:
`Can't find a binder for type: class java.time.LocalDate; `
Tracked a bit back and found that Spring Data fails to convert `java.time.LocalDate` to `com.google.cloud.Date`. Also possibly related, found this [LocalDateConverterOverride](https://github.com/spring-projects/spring-data-relational/blob/main/spring-data-r2dbc/src/main/java/org/springframework/data/r2dbc/convert/R2dbcConverters.java#L260-L269) in spring-data-r2dbc.

zhumin8, line 110: Unfortunately we don't run integration tests on external contributions, you will need to setup your own project to test this. 
I tried to run this with the comparison and constructor fixed, and find it fails with error:
`Can't find a binder for type: class java.time.LocalDate; `
Tracked a bit back and found that Spring Data fails to convert `java.time.LocalDate` to `com.google.cloud.Date`. Also possibly related, found this [LocalDateConverterOverride](https://github.com/spring-projects/spring-data-relational/blob/main/spring-data-r2dbc/src/main/java/org/springframework/data/r2dbc/convert/R2dbcConverters.java#L260-L269) in spring-data-r2dbc.

magiccrafter:
@zhumin8 I have defined the converters outside as custom ones, and it is working like a charm. I expected that it would work by adding them to the SpannerDialect. I will troubleshoot the code a bit to figure out the role of the overwriting converters and adapt the PR.

magiccrafter:
@zhumin8 @elefeint Spring is defining some very ambiguous date/time overrides which take precedence over the ones defined in the `SpannerR2dbcDialect ` class. I've raised a question with the maintainers of that spring data library, no answer so far. 

I see two options to proceed from here:
1. Bare minimum. 
Changes: 
    - Add only the missing google data types to the `getSimpleTypes()`:
```java
@Override
public Collection<? extends Class<?>> getSimpleTypes() {
  return Arrays.asList(
    JsonWrapper.class,
    Timestamp.class,
    ByteArray.class,
    Date.class);
}
 ```
 This bit is mandatory. Without that the `cloud-spanner-spring-data-r2dbc` cannot bind and convert from/to any of the 
 time java types and google custom types. If we choose to go with this option the client has to explicitly define every 
 read/write converter in a `R2dbcCustomConversions` bean. 

2. Complete.
Changes: 
    - All from Option 1.
    - Keep the custom read/write converters and define a `R2dbcCustomConversions` bean in a Spring Boot Auto-configuration bean:
    ```java
     @Bean
     @ConditionalOnMissingBean
     public R2dbcCustomConversions r2dbcCustomConversions() {
         return R2dbcCustomConversions.of(
             new SpannerR2dbcDialect(),
             Arrays.asList(
                 new DateToLocalDateConverter(),
                 new LocalDateToDateConverter(),
                 new LocalDateTimeToTimestampConverter(),
                 new TimestampToLocalDateTimeConverter()
             )
         );
     }
    ```
With this second approach, the `cloud-spanner-spring-data-r2dbc` lib will be usable out of the box without forcing the clients to write a custom converter for very basic date and time types. The bean is optional and allows overriding because of the `@ConditionalOnMissingBean` annotation. I've tested it in a project and it works like a charm.

Please let me know what you think and which direction you prefer. Here the question is more like this: 
Do you want to go with option 2?

elefeint:
Since this repository only has the driver and the dialect, we should not be adding Spring Boot autoconfiguration here. In the future, such convenience autoconfigurations will live in Spring Cloud GCP project, but we don't currently have a starter for Spanner R2DBC there.

For the purposes of this repo, document how a user would do (2) in the README -- that's what we've done for [JSON custom converters](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc#json-mapping) in the past.

PR #558, opened 2022-07-07
Upgrade spring boot and spring data

elefeint:
The individual dependabot upgrades seem to have failed because of a mismatch between autoconfigured Spring Boot objects and the functionality available.

I probably still want to track down where the reactive stream gets disconnected on runtime API mismatch (MethodNotFound or some such). It would be hard for users to troubleshoot.

### DIFF
--- a/cloud-spanner-r2dbc-samples/cloud-spanner-spring-data-r2dbc-sample/pom.xml
+++ b/cloud-spanner-r2dbc-samples/cloud-spanner-spring-data-r2dbc-sample/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
-    <version>2.6.8</version>
+    <version>2.7.1</version>
     <relativePath/>
   </parent>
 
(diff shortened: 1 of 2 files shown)


elefeint:
The underlying issue was our old friend, handling of `Exception` vs `Error`, but with a Project Reactor twist.
The root cause, an incompatibility between different Spring projects' versions, resulted in a runtime error `java.lang.NoSuchMethodError: 'java.lang.Object org.springframework.data.mapping.model.ParameterValueProvider.getParameterValue(org.springframework.data.mapping.Parameter)'`. 

Note two things: 
1) it's an `Error`, so not actually meant to be caught by anything 
2) it extends from `java.lang.LinkageError`, which project reactor [considers unrecoverable](https://projectreactor.io/docs/core/release/api/index.html?reactor/core/Exceptions.html#throwIfFatal). Which, fair. However, the presentation of all of this is that a flux subscription simply hangs.

Based on explanation in reactor/reactor-core#3036, this works as intended. This is understandable -- the exception ends up rethrown by Reactor layers, as Java intended. But because in our case, the caller is Spanner async API, the error ends up propagating there and also never getting logged.

What I think I'll do as follow-up is have a second `catch` block in `ReactiveResultSetCallback` that would handle `Throwable` (because there is no point handing it to Spanner async API), finish the async callback gracefully, and log the error. Unfortunately, the sink ends up being in a weird state, so the subscriber is still out of luck, but at least the error will be visible.

zhumin8 (approved):

Issue #609, opened 2022-12-07
Aggregate functions in cloud-spanner-spring-data-r2dbc repos throw IndexOutOfBoundsException


elmozgo:
I tried to create an example project with:

- **spring boot** (2.7.3)
- **cloud spanner**
- **r2dbc**
- **spring data**
- **liquibase**

I'm using **cloud-spanner-spring-data-r2dbc** version **1.1.0**

To be able to start the app, apply schema using liquibase and make basic findById() repository methods work, I had to downgrade the versions quite a bit. 

```
com.google.cloud->google-cloud-spanner-jdbc -> 2.4.4
com.google.cloudspannerecosystem:liquibase-spanner -> 4.9.1
io.r2dbc:r2dbc-spi -> 0.8.6.RELEASE
com.google.cloud:google-cloud-spanner -> 6.18.0
```

<details>
<summary>I created a simple entity:</summary>

```java
@Table("names")
public class NameEntity {

  @Id private String id;
  private String nickname;

  public String getId() {
    return id;
  }

  public void setId(String id) {
    this.id = id;
  }

  public String getNickname() {
    return nickname;
  }

  public void setNickname(String nickname) {
    this.nickname = nickname;
  }
}
```
</details>

<details> 
<summary>and a repo:</summary>

```java

public interface NamesRepository
    extends ReactiveCrudRepository<NameEntity, String>, NamesProvider {}

```

</details>

When testing the repository:

```java

class NamesRepositoryTest extends CloudSpannerEmulatorSupport {

  @Autowired private NamesRepository namesRepository;

  @Test
  void tryBasicRepoMethods() {

    StepVerifier.create(namesRepository.findById("a1a8687d-3206-47b6-99d9-1317b1de2c94")) // this works 
        .expectNextCount(1L)
        .verifyComplete();

    StepVerifier.create(namesRepository.findAll()).expectNextCount(5L).verifyComplete(); // this works 

    StepVerifier.create(namesRepository.count()).expectNext(5L).verifyComplete(); //this throws
  }
}

```

I get the exception:

```
...


Suppressed: java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 1
		at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
		at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
[...]

PR #615, opened 2022-12-14
fix: row column indexing in SpannerClientLibraryRow

meltsufin:
Fixes: #609

### DIFF
--- a/cloud-spanner-r2dbc-samples/cloud-spanner-spring-data-r2dbc-sample/src/main/java/com/example/Book.java
+++ b/cloud-spanner-r2dbc-samples/cloud-spanner-spring-data-r2dbc-sample/src/main/java/com/example/Book.java
@@ -16,9 +16,9 @@
 
 package com.example;
 
-import java.util.ArrayList;
 import java.util.List;
 import org.springframework.data.annotation.Id;
+import org.springframework.data.domain.Persistable;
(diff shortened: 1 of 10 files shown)


cloud-spanner-r2dbc-samples/cloud-spanner-spring-data-r2dbc-sample/src/main/java/com/example/Book.java
```diff
@@ -57,6 +57,11 @@ public String getId() {
     return id;
   }
 
+  @Override
+  public boolean isNew() {
+    return true;
```
meltsufin, line 62: This is to help Spring Data to know that this is a new object during `repository.save()` even though the ID is not null.

pom.xml
```diff
@@ -83,7 +83,6 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
-    <skip.surefire.tests>${skipTests}</skip.surefire.tests>
```
meltsufin, line 86: I believe this is redundant and it breaking IntelliJ. So, removing.

zhumin8 (approved):
LGTM

PR #738, opened 2023-10-11
feat: Implementing Connection#beginTransaction(TransactionDefinition) to support @Transactional annotation

### LINKED ISSUE #238: Implement the new extensible transaction definition in R2DBC SPI
Extend TransactionDefinition to support read-only, read-write, stale-read (and partitioned DML?) transactions with the  extensible transaction definition in  r2dbc/r2dbc-spi#159.

The transaction definition should map its own values to valid calls to already-implemented `DatabaseClientTransactionManager.beginReadonlyTransaction(TimestampBound timestampBound)`

jainsahab:
Implemented the following classes and methods to conform to R2DBC SPI which enables supporting `@Transactional` annotation in [Spring Data Dialect](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/tree/main/spring-cloud-spanner-spring-data-r2dbc).

- `Connection#beginTransaction(TransactionDefinition)` method so that it can be used by a `TransactionManager` when `@Transactional` annotation is used. ✅
- `Connection#getTransactionIsolationLevel` returns SERIALIZABLE as it is the closest level to EXTERNAL_CONSISTENCY which spanner supports. ✅
- `Connection#setTransactionIsolationLevel(IsolationLevel)` accepts only `IsolationLevel.SERIALIZABLE` and throw `UnsupportedOperationException` otherwise. ✅
- Introduced `SpannerTransactionDefinition`, an implementation of `TransactionDefinition`(R2DBC spi) to configure the transaction attributes which will be used by `Connection#beginTransaction(TransactionDefinition)` method. ✅

Fixes #238

### DIFF
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
@@ -58,7 +58,9 @@ class SpannerClientLibraryConnection implements Connection, SpannerConnection {
 
   @Override
   public Publisher<Void> beginTransaction(TransactionDefinition definition) {
-    return validateIsolation(definition.getAttribute(ISOLATION_LEVEL))
+    IsolationLevel isolationLevel = firstNonNull(definition.getAttribute(ISOLATION_LEVEL),
+        SERIALIZABLE);
+    return validateIsolation(isolationLevel)
         .then(Mono.defer(() -> {
           boolean isReadOnly = TRUE.equals(definition.getAttribute(READ_ONLY));
           if (isReadOnly) {
@@ -154,12 +156,9 @@ class SpannerClientLibraryConnection implements Connection, SpannerConnection {
     return this.clientLibraryAdapter.setAutoCommit(autoCommit);
   }
 
+
   @Override
   public Publisher<Void> setTransactionIsolationLevel(IsolationLevel isolationLevel) {
-    if (isolationLevel == null) {
-      return Mono.error(new IllegalArgumentException("IsolationLevel can't be null."));
-    }
-
     return this.validateIsolation(isolationLevel);
   }
 
@@ -182,8 +181,11 @@ class SpannerClientLibraryConnection implements Connection, SpannerConnection {
   }
 
   private Mono<Void> validateIsolation(IsolationLevel isolationLevel) {
-    boolean valid = isolationLevel == null || isolationLevel == SERIALIZABLE;
-    return valid ? Mono.empty() : Mono.error(new UnsupportedOperationException(
-        String.format("'%s' isolation level not supported", isolationLevel.asSql())));
+    if (isolationLevel == null) {
+      return Mono.error(new IllegalArgumentException("IsolationLevel can't be null."));
+    }
+    return isolationLevel == SERIALIZABLE ? Mono.empty()
+        : Mono.error(new UnsupportedOperationException(
+            String.format("'%s' isolation level not supported", isolationLevel.asSql())));
   }
 }
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
@@ -183,7 +183,7 @@ class SpannerClientLibraryConnection implements Connection, SpannerConnection {
 
   private Mono<Void> validateIsolation(IsolationLevel isolationLevel) {
     boolean valid = isolationLevel == null || isolationLevel == SERIALIZABLE;
-    return valid ? Mono.empty(): Mono.error(new UnsupportedOperationException(
+    return valid ? Mono.empty() : Mono.error(new UnsupportedOperationException(
         String.format("'%s' isolation level not supported", isolationLevel.asSql())));
   }
 }
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
@@ -43,6 +43,9 @@ import org.reactivestreams.Publisher;
 import reactor.core.publisher.Mono;
 
 class SpannerClientLibraryConnection implements Connection, SpannerConnection {
+  private final static List<IsolationLevel> UNSUPPORTED_ISOLATION_LEVELS = Arrays.asList(READ_COMMITTED,
+      READ_UNCOMMITTED,
+      REPEATABLE_READ);
 
   private final DatabaseClientReactiveAdapter clientLibraryAdapter;
 
@@ -160,6 +163,10 @@ class SpannerClientLibraryConnection implements Connection, SpannerConnection {
 
   @Override
   public Publisher<Void> setTransactionIsolationLevel(IsolationLevel isolationLevel) {
+    if (isolationLevel == null) {
+      return Mono.error(new IllegalArgumentException("IsolationLevel can't be null."));
+    }
+
     return this.validateIsolation(isolationLevel);
   }
 
@@ -182,14 +189,11 @@ class SpannerClientLibraryConnection implements Connection, SpannerConnection {
   }
 
   private Mono<Void> validateIsolation(IsolationLevel isolationLevel) {
-    List<IsolationLevel> unsupportedIsolationLevels = Arrays.asList(READ_COMMITTED,
-        READ_UNCOMMITTED,
-        REPEATABLE_READ);
-    boolean valid = !unsupportedIsolationLevels.contains(isolationLevel);
-    if (valid) {
-      return Mono.empty();
+    boolean invalid = UNSUPPORTED_ISOLATION_LEVELS.contains(isolationLevel);
+    if (invalid) {
+      return Mono.error(new UnsupportedOperationException(
+          String.format("%s isolation level not supported", isolationLevel)));
     }
-    return Mono.error(new UnsupportedOperationException(
-        String.format("%s isolation level not supported", isolationLevel)));
+    return Mono.empty();
   }
 }
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
@@ -186,6 +186,7 @@ class SpannerClientLibraryConnection implements Connection, SpannerConnection {
     }
     return isolationLevel == SERIALIZABLE ? Mono.empty()
         : Mono.error(new UnsupportedOperationException(
-            String.format("'%s' isolation level not supported", isolationLevel.asSql())));
+            String.format("Unsupported '%s' isolation level, Only SERIALIZABLE is supported.",
+                isolationLevel.asSql())));
   }
 }
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
@@ -16,6 +16,15 @@
 
 package com.google.cloud.spanner.r2dbc.v2;
 
+import static com.google.cloud.spanner.r2dbc.v2.SpannerConstants.TIMESTAMP_BOUND;
+import static com.google.common.base.MoreObjects.firstNonNull;
+import static io.r2dbc.spi.IsolationLevel.READ_COMMITTED;
+import static io.r2dbc.spi.IsolationLevel.READ_UNCOMMITTED;
+import static io.r2dbc.spi.IsolationLevel.REPEATABLE_READ;
+import static io.r2dbc.spi.TransactionDefinition.ISOLATION_LEVEL;
+import static io.r2dbc.spi.TransactionDefinition.READ_ONLY;
+import static java.lang.Boolean.TRUE;
+
 import com.google.cloud.spanner.TimestampBound;
 import com.google.cloud.spanner.r2dbc.api.SpannerConnection;
 import com.google.cloud.spanner.r2dbc.statement.StatementParser;
@@ -28,6 +37,8 @@ import io.r2dbc.spi.Statement;
 import io.r2dbc.spi.TransactionDefinition;
 import io.r2dbc.spi.ValidationDepth;
 import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Mono;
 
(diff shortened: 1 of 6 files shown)


cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnectionTest.java
```diff
@@ -60,6 +67,127 @@ void beginReadonlyTransactionUsesStrongConsistencyByDefault() {
...
+
+    verify(this.mockAdapter).beginTransaction();
+  }
+
+  @Test
+  void shouldBeginTransactionInReadWriteModeByDefault() {
+    SpannerTransactionDefinition readWriteDefinition = new SpannerTransactionDefinition.Builder()
+        .build();   // absence of attribute indicates read write transaction
+
+    when(this.mockAdapter.beginTransaction()).thenReturn(Mono.empty());
```
burkedavison, line 103: Variations of this "mock so it doesn't explode" line are repeated in many of these tests.

If you want to simplify things a bit, consider adding the following to the @BeforeEach method:
```java
    when(this.mockAdapter.beginReadonlyTransaction(any())).thenReturn(Mono.empty());
    when(this.mockAdapter.beginTransaction()).thenReturn(Mono.empty());
```

Then you can delete all the equivalent happy-path mock setups in each `@Test` including this one:
```java 
    when(this.mockAdapter.beginReadonlyTransaction(fiveSecondTimestampBound))
        .thenReturn(Mono.empty());
```

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerConstants.java
```diff
@@ -0,0 +1,28 @@
...
+package com.google.cloud.spanner.r2dbc.v2;
+
+import com.google.cloud.spanner.TimestampBound;
+import io.r2dbc.spi.Option;
+
+/**
+ * Spanner Constants.
+ */
+public class SpannerConstants {
+  public static Option<TimestampBound> TIMESTAMP_BOUND = Option.valueOf("timestampBound");
```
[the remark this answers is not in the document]
jainsahab: Done, fixed in 45673cf.

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnectionTest.java
jainsahab: Thanks for pointing it out @burkedavison , fixed in 45673cf.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
```diff
@@ -28,11 +38,16 @@
 import io.r2dbc.spi.TransactionDefinition;
 import io.r2dbc.spi.ValidationDepth;
 import java.time.Duration;
+import java.util.List;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Mono;
 
 class SpannerClientLibraryConnection implements Connection, SpannerConnection {
 
+  private static final List<IsolationLevel> UNSUPPORTED_ISOLATION_LEVELS = asList(READ_COMMITTED,
```
meltsufin, line 47: Why not have an allowlist instead of a blocklist?

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
```diff
@@ -51,7 +66,16 @@ public Publisher<Void> beginTransaction() {
 
   @Override
   public Publisher<Void> beginTransaction(TransactionDefinition definition) {
-    return Mono.error(new UnsupportedOperationException());
+    return validateIsolation(definition.getAttribute(ISOLATION_LEVEL))
```
meltsufin, line 69: Is this resolving #238?

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
```diff
@@ -140,7 +164,11 @@ public Publisher<Void> setAutoCommit(boolean autoCommit) {
 
   @Override
   public Publisher<Void> setTransactionIsolationLevel(IsolationLevel isolationLevel) {
-    throw new UnsupportedOperationException();
+    if (isolationLevel == null) {
+      return Mono.error(new IllegalArgumentException("IsolationLevel can't be null."));
+    }
+
+    return this.validateIsolation(isolationLevel);
```
meltsufin, line 171: I don't think we need this check. Serializable is the strongest isolation level. So, it's essentially a superset of other weaker isolation levels. If someone wants read-uncommitted, they get it with serializable.
@olavloite WDYT?

olavloite: We had that discussion for a couple of other drivers as well, and landed on the conclusion that we want to be strict here. The reason is that:
1. Cloud Spanner might introduce new isolation levels in the future.
2. We would then want to be able to plug those isolation levels in here.
3. Accepting for example REPEATABLE_READ here now, while actually using SERIALIZABLE, would mean that changing REPEATABLE_READ to actually use a different isolation level in the future will be a breaking change. That means that any new isolation level will require a major version bump of this driver.

I'm not strictly opposed to being less strict here, but we should at least understand the consequences of it. So far, we've landed on the 'only allow serializable'-option for other drivers.

meltsufin: Agreed, for consistency sake and future-proofing.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
jainsahab: Partially yes,

#238 talks about  read-only, read-write, stale-read and partitioned DML. Out of these first three are supported but not partitioned DML but that is not supported by the driver itself.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
jainsahab: There is no reason as such, but I've simplified the validations in 0bbca1c.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
meltsufin: Can we create a separate issue for DML support, and close #238 as part of this PR?

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
```diff
@@ -140,7 +156,11 @@ public Publisher<Void> setAutoCommit(boolean autoCommit) {
 
   @Override
   public Publisher<Void> setTransactionIsolationLevel(IsolationLevel isolationLevel) {
-    throw new UnsupportedOperationException();
+    if (isolationLevel == null) {
+      return Mono.error(new IllegalArgumentException("IsolationLevel can't be null."));
+    }
```
meltsufin, line 161: I would consider this part of validating the `isolationLevel`. Can we just move this `if` in to the method below.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
```diff
@@ -160,4 +180,10 @@ public Publisher<Void> close() {
   public boolean isInReadonlyTransaction() {
     return this.clientLibraryAdapter.isInReadonlyTransaction();
   }
+
+  private Mono<Void> validateIsolation(IsolationLevel isolationLevel) {
+    boolean valid = isolationLevel == null || isolationLevel == SERIALIZABLE;
```
meltsufin, line 185: Why would `null` be OK?

jainsahab: > Why would `null` be OK?

When you use `@Transactional` annotation without specifying the `isolation` parameter, It's a valid use case, and library ought to use default Isolation in that case,

jainsahab: I have addressed the confusion  now.

So `validateIsolation` will always throw exception if a null value is passed which is a requirement from SPI when `setTransactionIsolationLevel` is invoked with a null parameter.

when `validateIsolation` is called from `beginTransaction(TransactionDefinition)`, it will never be null as we will assume it to be `SERIALIZABLE` in that case.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
jainsahab: Sounds good to me, I've created #747.

and whenever you close this issue, it'll will close #238

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
jainsahab: Done

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
```diff
@@ -160,4 +179,13 @@ public Publisher<Void> close() {
...
     return this.clientLibraryAdapter.isInReadonlyTransaction();
   }
+
+  private Mono<Void> validateIsolation(IsolationLevel isolationLevel) {
+    if (isolationLevel == null) {
+      return Mono.error(new IllegalArgumentException("IsolationLevel can't be null."));
+    }
+    return isolationLevel == SERIALIZABLE ? Mono.empty()
+        : Mono.error(new UnsupportedOperationException(
+            String.format("'%s' isolation level not supported", isolationLevel.asSql())));
```
olavloite, line 189: nit: Maybe add to the error message that only `SERIALIZABLE` is supported.

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnectionTest.java
```diff
@@ -47,19 +54,130 @@ class SpannerClientLibraryConnectionTest {
...
+        .verifyComplete();
+
+    verify(this.mockAdapter).beginTransaction();
+  }
+
+  @Test
+  void shouldBeginTransactionWithGivenTimestampBound() {
+    TimestampBound fiveSecondTimestampBound = TimestampBound.ofExactStaleness(5L, TimeUnit.SECONDS);
+
+    SpannerTransactionDefinition readWriteDefinition = new SpannerTransactionDefinition.Builder()
```
olavloite, line 109: nit:
```suggestion
    SpannerTransactionDefinition staleTransactionDefinition = new SpannerTransactionDefinition.Builder()
```

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
```diff
@@ -51,7 +58,18 @@ public Publisher<Void> beginTransaction() {
...
+        SERIALIZABLE);
+    return validateIsolation(isolationLevel)
+        .then(Mono.defer(() -> {
+          boolean isReadOnly = TRUE.equals(definition.getAttribute(READ_ONLY));
+          if (isReadOnly) {
+            TimestampBound timestampBound = firstNonNull(definition.getAttribute(TIMESTAMP_BOUND),
+                TimestampBound.strong());
+            return this.clientLibraryAdapter.beginReadonlyTransaction(timestampBound);
+          }
+          return this.clientLibraryAdapter.beginTransaction();
```
olavloite, line 71: I think that we should throw an error if the user has created a `TransactionDefinition` for a read/write transaction with a read-only staleness. That is; if they have set a `TIMESTAMP_BOUND` option, but not set `READ_ONLY` to true.

jainsahab: Addressed in [4e41ba2](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/738/commits/4e41ba2f4fc1c632c74b4ead8dacbcb7d2934961).

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnection.java
jainsahab: Addressed in [4e41ba2](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/738/commits/4e41ba2f4fc1c632c74b4ead8dacbcb7d2934961).

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/v2/SpannerClientLibraryConnectionTest.java
jainsahab: couldn't accept this suggestion, as this variable was being used below this line and had to be changed there as well, fixed in [4e41ba2](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/pull/738/commits/4e41ba2f4fc1c632c74b4ead8dacbcb7d2934961).

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerTransactionDefinition.java
```diff
@@ -29,9 +32,17 @@ public class SpannerTransactionDefinition implements TransactionDefinition {
   private final Map<Option<?>, Object> internalMap;
 
   SpannerTransactionDefinition(Map<Option<?>, Object> internalMap) {
+    validate(internalMap);
     this.internalMap = internalMap;
   }
 
+  private void validate(Map<Option<?>, Object> internalMap) {
+    if (FALSE.equals(internalMap.get(READ_ONLY)) && internalMap.containsKey(TIMESTAMP_BOUND)) {
```
olavloite, line 40: What happens if the user has not set any value for `READ_ONLY`? The default is then read/write, and the `TIMESTAMP_BOUND` will be ignored. So should the test not also include a check whether there is no `READ_ONLY` option?
(And maybe centralize that in a single, simple method named something like `isReadOnlyTransaction()`

olavloite (approved):
LGTM, with one tiny nit on the last check for using a stale read option with a (default) read/write transaction.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/SpannerTransactionDefinition.java
jainsahab: My bad 😓 , fixed that too now.

PR #888, opened 2025-04-30
fix: rollback could return session to pool twice

olavloite:
Transaction rollbacks could cause a session to be added to the session pool twice. This again could cause two transactions to try to use the same session at the same time, which again could lead to unpredictable errors.

### DIFF
--- a/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
+++ b/cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
@@ -20,6 +20,7 @@ import com.google.auth.oauth2.GoogleCredentials;
 import com.google.auth.oauth2.OAuth2Credentials;
 import com.google.cloud.spanner.SpannerOptions;
 import com.google.cloud.spanner.r2dbc.util.Assert;
+import io.grpc.ManagedChannelBuilder;
 import io.r2dbc.spi.ConnectionFactoryOptions;
 import io.r2dbc.spi.R2dbcNonTransientResourceException;
 import java.io.IOException;
@@ -51,11 +52,13 @@ public class SpannerConnectionConfiguration {
   // TODO: check how to handle full URL (it gets parsed by SPI, we only get pieces)
   private final String fullyQualifiedDbName;
 
-  private String projectId;
+  private final ConnectionFactoryOptions options;
 
-  private String instanceName;
+  private final String projectId;
 
-  private String databaseName;
+  private final String instanceName;
+
+  private final String databaseName;
 
   private final OAuth2Credentials credentials;
 
@@ -82,6 +85,7 @@ public class SpannerConnectionConfiguration {
    * @param credentials GCP credentials to authenticate service calls with.
    */
   private SpannerConnectionConfiguration(
+      ConnectionFactoryOptions options,
       String projectId,
       String instanceName,
       String databaseName,
@@ -91,6 +95,7 @@ public class SpannerConnectionConfiguration {
     Assert.requireNonNull(instanceName, "instanceName must not be null");
     Assert.requireNonNull(databaseName, "databaseName must not be null");
 
+    this.options = Assert.requireNonNull(options, "options must not be null");
     this.projectId = projectId;
     this.instanceName = instanceName;
     this.databaseName = databaseName;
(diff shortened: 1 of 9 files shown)


cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -51,11 +52,13 @@ public class SpannerConnectionConfiguration {
   // TODO: check how to handle full URL (it gets parsed by SPI, we only get pieces)
   private final String fullyQualifiedDbName;
 
-  private String projectId;
+  private final ConnectionFactoryOptions options;
```
olavloite, line 55: We're adding the standard R2DBC `ConnectionFactoryOptions` to this class, so we can read them when building the `SpannerOptions` to use. This allows us to connect to a custom server, such as for example an in-mem mock server for testing.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -189,13 +194,24 @@ public int hashCode() {
...
 
+    if (this.options.hasOption(ConnectionFactoryOptions.HOST)) {
+      int port = 443;
+      if (this.options.hasOption(ConnectionFactoryOptions.PORT)) {
+        port = (Integer) Objects.requireNonNull(
+            this.options.getValue(ConnectionFactoryOptions.PORT));
+      }
+      String host = String.format("http://%s:%d", this.options.getValue(ConnectionFactoryOptions.HOST), port);
+      optionsBuilder.setHost(host);
+    }
```
olavloite, line 205: HOST and PORT options were previously ignored. We need it for this change, as it also adds testing against an in-mem mock server, and we need to be able to connect to the randomly assigned address of that server.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/SpannerConnectionConfiguration.java
```diff
@@ -189,13 +194,24 @@ public int hashCode() {
...
     if (this.projectId != null) {
       optionsBuilder.setProjectId(this.projectId);
     }
-
     if (this.credentials != null) {
       optionsBuilder.setCredentials(this.credentials);
     }
+    if (this.usePlainText) {
+      optionsBuilder.setChannelConfigurator(ManagedChannelBuilder::usePlaintext);
+    }
```
olavloite, line 214: This option was already there, but was not used.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientTransactionManager.java
```diff
@@ -108,7 +109,9 @@ ApiFuture<Void> clearTransactionManager() {
     ApiFuture<Void> returnFuture = ApiFutures.immediateFuture(null);
 
     if (this.transactionManager != null) {
-      returnFuture = this.transactionManager.closeAsync();
+      if (this.transactionManager.getState() != TransactionState.ROLLED_BACK) {
+        returnFuture = this.transactionManager.closeAsync();
+      }
```
olavloite, line 114: This is the actual fix in this case. Closing a TransactionManager that has already been rolled back, could cause the underlying session to be added to the session pool twice. This will also be fixed in the Spanner client library, making a second call to close a no-op.

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/SpannerConnectionFactoryProviderTest.java
```diff
@@ -101,6 +103,8 @@ void testCreate() {
         this.spannerConnectionFactoryProvider.create(SPANNER_OPTIONS);
     assertThat(spannerConnectionFactory).isNotNull();
     assertThat(spannerConnectionFactory).isInstanceOf(SpannerClientLibraryConnectionFactory.class);
+
+    Flux.from(((Closeable) spannerConnectionFactory).close()).blockLast();
```
olavloite, line 107: These close calls have been added to prevent gRPC channels that are created when a `SpannerConnectionFactory` is created to remain open. If this is not added, then the log is full of warnings about unclosed channels.

cloud-spanner-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/v2/MockServerTest.java
```diff
@@ -0,0 +1,310 @@
...
+              .blockLast();
+        });
+      });
+      Future<?> queryFuture = queryService.submit(() -> {
+        repeat(n, () -> {
+          Flux.concat(connection2.createStatement(sql).execute()).blockLast();
+        });
+      });
+      rollbackFuture.get();
+      queryFuture.get();
```
olavloite, line 280: This would fail without the fix in this PR. Rolling back an (empty) transaction while at the same time running a bunch of other statements/transactions, could cause two of them to use the same session, and then one transaction would invalidate the other, causing one of them to fail.

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientTransactionManager.java
jinseopkim0: Thank you for the explanation. Could we add a brief comment why we need this? E.g.
// Prevent closing a TransactionManager that has already been rolled back, to avoid returning the session to the session pool twice.

Also, are there other TransactionState that could be good to check? Maybe we should also avoid calling for ABORTED?

jinseopkim0 (changes requested):
Thank you for the PR!

cloud-spanner-r2dbc/src/main/java/com/google/cloud/spanner/r2dbc/v2/DatabaseClientTransactionManager.java
olavloite: I've added an explanation to the code.

There are no other transaction states where we should skip the close call. I've verified it manually by looking into whether the session is being returned twice, but it's hard to create a test shows it. Also; if it had been the case, then we would have seen this error much more frequently.

jinseopkim0 (approved):

